PR #59, opened 2018-04-07
Removal of Ruby dependancy for ITT map script compilation.

crojewsk:
Disclaimer:
Objective of this change is to remove ruby dependancy while keeping code is status quo.

- Implemented compile.ps1 powershell script file which goal is to complite .j and .zn files using jasshelper.
- Updated .gitignore to account for new generated file.
- Converted all .erb files to pure .j or .zn files.
- Fixed file extension for TypeStructs file.
- Moved MapOut.j outside of /init directory to keep its content at the end of generated file as it should be.
- Added CHEATS_ENABLED global boolean for controlling behavior of CheatPack library initialization.


compile.ps1
```diff
@@ -0,0 +1,50 @@
...
+    }
+    Else
+    {
+        Get-Content -Path $File | Add-Content $GeneratedJ
+    }
+}
+
+# Run jasshelper to perform actual complitaion of map script
+$Params = @('--scriptonly', $JasshelperArgs, $CommonJ, $BlizzardJ, $GeneratedJ, $OutputJ)
+& $Jasshelper $Params
```
theQuazz: Really nice work, this is awesome! So much cleaner than the ruby version haha.

jonathanmclaus:
Did you look at the `.vscode/tasks.json` file? I think that needs to be changed to keep the VS Code plugin working.

jass/lib/CompileInfo.zn
```diff
@@ -2,15 +2,15 @@ library CompileInfo requires Constants, ArgsUtils {
...
+    static constant integer year       = 1;//<%= Time.now.strftime("%Y").to_i %>;
+    static constant integer dayOfYear  = 1;//<%= Time.now.strftime("%j").to_i %>;
+    static constant integer month      = 1;//<%= Time.now.strftime("%m").to_i %>;
+    static constant integer day        = 1;//<%= Time.now.strftime("%d").to_i %>;
+    static constant integer hours      = 1;//<%= Time.now.strftime("%H").to_i %>;
+    static constant integer minutes    = 1;//<%= Time.now.strftime("%M").to_i %>;
+    static constant integer seconds    = 1;//<%= Time.now.strftime("%S").to_i %>;
 
     static method operator dateTime () -> string {
       return I2S(year) + "-" + I2S(month) + "-" + I2S(day) + " " +
```
theQuazz: My feeling is that we can just remove this file. I'm not sure its actually doing anything important.

jass/lib/Map.zn
```diff
@@ -6,7 +6,7 @@ library Map {
     static integer ENV_BETA = 102;
     static integer ENV_TEST = 103;
 
-    private static integer currentEnv = ENV_<%= @opt[:env].upcase %>;
+    private static integer currentEnv = 1; //ENV_<%= @opt[:env].upcase %>;
 
     static method getEnv() -> integer {
       return currentEnv;
```
theQuazz: I think we should remove remove lines 5-13 in this file.

theQuazz:
To @jonathanmclaus' point, perhaps we should remove the compile script written in ruby and link the build task to this file.

crojewsk:
Wurst build task works just fine.
Please read the disclaimer: goal of this change is to remove ruby dependancy while keep code in status quo. Ruby-compile script will be removed later.
All classes converted to pure .j or .zn will be marked OBSOLETE and are scheduled for removal, that is why I did not bother with any other changes.

theQuazz:
@crojewsk sounds good, merge when you're ready!

crojewsk:
I do not have write permission : D

PR #60, opened 2018-04-07
 Replace Cohadar's jasshelper with Vexorian's.

crojewsk:
Status of code compilation: unchanged.
One can find the comparison of jasshelper verions on wc3c or hiveworkshop site.

Due to Cohadar leaving his work unfinished, his jasshelper was never accepted officialy and was left as optional in certain versions of JNPG while Vex's was always a default one.

For SharpCraft World Editor, Vexorian jasshelper is the only available and it is in best interest of ITT development to stick to official solution.


crojewsk:
Hmm 1st commit has been appended unexpectedly. Give me a sec.

PR #72, opened 2018-04-20
Addition of new item-related packages.

crojewsk:
Listed packages are added to replace existing ITT item handling.
Further patches will introduce usage of new and removal of obsolete
code.

- InventoryEvent, for intuivite inventory event handling
- ItemRecipe, powerful item recipe creator
- ItemRestriction, for restricting or limiting items from being equipped
- SmoothItemPickup, allows for item pickup during certain conditions
even when unit inventory is full
- StackNSplit, easy item charges stacking and splitting

### DIFF
--- a/wurst/lib/InventoryEvent.wurst
+++ b/wurst/lib/InventoryEvent.wurst
@@ -1,0 +1,98 @@
+/*
+*  InventoryEvent v1.0.0.0
+*     by Bannar
+*
+*  For intuitive inventory event handling.
+*/
+package InventoryEvent
+import RegisterEvents
+
+unit eventUnit = null
+item eventItem = null
+int eventSlotFrom = -1
+int eventSlotTo = -1
+trigger eventMovedTrigger = CreateTrigger()
+trigger eventUsedTrigger = CreateTrigger()
+
+public constant EVENT_INVENTORY_MOVED = 0
+public constant EVENT_INVENTORY_USED = 1
+
+public function getInventoryManipulatingUnit() returns unit
+    return eventUnit
+
+public function getInventoryManipulatedItem() returns item
+    return eventItem
+
+public function getInventorySlotFrom() returns int
+    return eventSlotFrom
+
+public function getInventorySlotTo() returns int
+    return eventSlotTo
+
+public function getInventorySwappedItem() returns item
+    return UnitItemInSlot(eventUnit, eventSlotTo)
+
+public function unit.getItemSlot(item itm) returns int
+    if this.hasItem(itm)
+        let size = this.inventorySize()
+        for i = 0 to size-1
+            if this.itemInSlot(i) == itm
+                return i
+    return -1
+
+function fireEvent(trigger evt, unit u, item itm, integer slotFrom, integer slotTo)
+    unit prevUnit = eventUnit
+    item prevItem = eventItem
+    int prevSlotFrom = eventSlotFrom
+    int prevSlotTo = eventSlotTo
+
+    eventUnit = u
+    eventItem = itm
+    eventSlotFrom = slotFrom
+    eventSlotTo = slotTo
+
+    evt.evaluate()
+
+    eventUnit = prevUnit
+    eventItem = prevItem
+    eventSlotFrom = prevSlotFrom
+    eventSlotTo = prevSlotTo
+
+function onItemOrder()
+    let order = GetIssuedOrderId()
+    let u = GetTriggerUnit()
+    item itm
+    int slotFrom
+    int slotTo
+
+    if order >= 852002 and order <= 852007
+        itm = GetOrderTargetItem()
+        slotFrom = u.getItemSlot(itm)
+        slotTo = order - 852002
+        fireEvent(eventMovedTrigger, u, itm, slotFrom, slotTo)
+    else
+        slotFrom = order - 852008
+        itm = u.itemInSlot(slotFrom)
+        fireEvent(eventUsedTrigger, u, itm, slotFrom, -1)
+
+function onAnyOrder()
+    let order = GetIssuedOrderId()
+    if order >= 852002 and order <= 852013
+        onItemOrder()
+
+public function registerInventoryEvent(int ev, code c)
+    if ev == EVENT_INVENTORY_MOVED
+        eventMovedTrigger.addCondition(Condition(c))
+    else
+        eventUsedTrigger.addCondition(Condition(c))
+
+public function getInventoryEventTrigger(int whichEvent) returns trigger
+    if whichEvent == EVENT_INVENTORY_MOVED
+        return eventMovedTrigger
+    return eventUsedTrigger
+
+init
+    // MOVED is order of type TARGET_ORDER yet USED can be anyone of them
+    registerPlayerUnitEvent(EVENT_PLAYER_UNIT_ISSUED_ORDER, () -> onAnyOrder())
+    registerPlayerUnitEvent(EVENT_PLAYER_UNIT_ISSUED_TARGET_ORDER, () -> onAnyOrder())
+    registerPlayerUnitEvent(EVENT_PLAYER_UNIT_ISSUED_POINT_ORDER, () -> onAnyOrder())
--- a/wurst/lib/InventoryEvent.wurst
+++ b/wurst/lib/InventoryEvent.wurst
@@ -32,10 +32,23 @@ public function getInventorySlotTo() returns int
 public function getInventorySwappedItem() returns item
     return UnitItemInSlot(eventUnit, eventSlotTo)
 
+public function registerInventoryEvent(int whichEvent, code cb)
+    if whichEvent == EVENT_INVENTORY_MOVED
+        eventMovedTrigger.addCondition(Condition(cb))
+    else if whichEvent == EVENT_INVENTORY_USED
+        eventUsedTrigger.addCondition(Condition(cb))
+
+public function getInventoryEventTrigger(int whichEvent) returns trigger
+    if whichEvent == EVENT_INVENTORY_MOVED
+        return eventMovedTrigger
+    else if whichEvent == EVENT_INVENTORY_USED
+        return eventUsedTrigger
+    return null
+
 public function unit.getItemSlot(item itm) returns int
     if this.hasItem(itm)
         let size = this.inventorySize()
-        for i = 0 to size-1
+        for i = 0 to size - 1
             if this.itemInSlot(i) == itm
                 return i
     return -1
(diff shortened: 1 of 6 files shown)


wurst/lib/InventoryEvent.wurst
```diff
@@ -0,0 +1,98 @@
...
+    else
+        eventUsedTrigger.addCondition(Condition(c))
+
+public function getInventoryEventTrigger(int whichEvent) returns trigger
+    if whichEvent == EVENT_INVENTORY_MOVED
+        return eventMovedTrigger
+    return eventUsedTrigger
+
+init
+    // MOVED is order of type TARGET_ORDER yet USED can be anyone of them
```
jonathanmclaus: Not too concerned with it currently, but let's consistently use this style of comment for at least single line ones.

wurst/lib/InventoryEvent.wurst
```diff
@@ -0,0 +1,98 @@
...
+    return eventSlotFrom
+
+public function getInventorySlotTo() returns int
+    return eventSlotTo
+
+public function getInventorySwappedItem() returns item
+    return UnitItemInSlot(eventUnit, eventSlotTo)
+
+public function unit.getItemSlot(item itm) returns int
+    if this.hasItem(itm)
```
jonathanmclaus: What's the efficiency on this like? I'm guessing that `hasItem` checks each slot in turn, so this is checking slots twice. I'd just remove this `if` statement and use the rest of this function as-is (after removing indent).

wurst/lib/InventoryEvent.wurst
```diff
@@ -0,0 +1,98 @@
...
+        slotTo = order - 852002
+        fireEvent(eventMovedTrigger, u, itm, slotFrom, slotTo)
+    else
+        slotFrom = order - 852008
+        itm = u.itemInSlot(slotFrom)
+        fireEvent(eventUsedTrigger, u, itm, slotFrom, -1)
+
+function onAnyOrder()
+    let order = GetIssuedOrderId()
+    if order >= 852002 and order <= 852013
```
jonathanmclaus: Please either comment magic numbers or place them into descriptive variables (possibly both for particularly complicated ones).

wurst/lib/InventoryEvent.wurst
```diff
@@ -0,0 +1,98 @@
...
+        onItemOrder()
+
+public function registerInventoryEvent(int ev, code c)
+    if ev == EVENT_INVENTORY_MOVED
+        eventMovedTrigger.addCondition(Condition(c))
+    else
+        eventUsedTrigger.addCondition(Condition(c))
+
+public function getInventoryEventTrigger(int whichEvent) returns trigger
+    if whichEvent == EVENT_INVENTORY_MOVED
```
jonathanmclaus: I'd be consistent with the style you used for `registerInventoryEvent`, i.e. if-else.

wurst/lib/ItemRecipe.wurst
```diff
@@ -0,0 +1,527 @@
...
+    function getIngredients() returns LinkedList<ItemData>
+        return new LinkedList<ItemData>(ingredients)
+
+    /** Returns item type of this recipe reward. */
+    function getResult() returns int
+        return reward
+
+    /** Starts single-reference counted batch. Allows to assign multiple items to the same item slot. */
+    function startBatch()
+        if not batch
```
jonathanmclaus: Currently this check doesn't do anything. Either way, `batch` is `true` afterwards so simply assigning it is sufficient. If you plan on adding an `else` branch, such as for error logging, please add a `TODO` comment for that.

wurst/lib/ItemRecipe.wurst
```diff
@@ -0,0 +1,527 @@
...
+        this.charges = base.charges
+        this.index = base.index
+
+/** Item recipe, collection of items that can be combined for powerful rewards. */
+public class ItemRecipe
+    private LinkedList<ItemData> ingredients
+    private int reward
+    int charges
+    boolean ordered
+    boolean permanent
```
jonathanmclaus: I think that we should collectively decide on a convention for booleans. I prefer `is` as a prefix, e.g. `isOrdered` and `isPermanent`, although `should` fits better in some cases.

wurst/lib/ItemRestriction.wurst
```diff
@@ -0,0 +1,393 @@
...
+    return "This item can not be hold by this unit type."
+
+@configurable function getLevelErrorMessage(int level) returns string
+    return "This item requires level " + level.toString() + " to be picked up."
+
+@configurable function getStatisticErrorMessage(int value, string statistic) returns string
+    return "This item requires " + value.toString() + " " + statistic + "."
+
+@configurable function getLimitErrorMessage(int limit) returns string
+    return "This unit can not hold more than " + limit.toString() + " items of this item type."
```
jonathanmclaus: Minor nitpick, but the `s` in `items` will need to be dropped if `limit == 1`.

wurst/lib/StackNSplit.wurst
```diff
@@ -0,0 +1,241 @@
...
+    int curr
+    int diff
+    int size = u.inventorySize()
+    item with
+
+    if charges == 0 or not isItemStackable(itemTypeId)
+        return false
+
+    int max = getItemMaxStacks(itemTypeId)
+    for i = 0 to size-1
```
jonathanmclaus: I don't think it makes sense to omit spaces around binary operators in control statements but not assignments, e.g. `diff = max - curr`. Not too concerned with this though.

wurst/lib/InventoryEvent.wurst
crojewsk: I plan to keep '//' for single, and '/** */' for descriptions and '/* */' for headers. This is also true for wurstStdlib2.

wurst/lib/InventoryEvent.wurst
crojewsk: If that was the case, imaged what IsItemOwned would do. Iterate over units on entire map and check them out?
Nope, it's almost always better to call those before proceeding with jass operations. Jass is tremendously slower than c++, and I believe they store references to owner of given item, thus the call is light-quick.

pseudo code:
bool struct_item::isItemOwned(struct_unit u) { return m_owner != null; }
bool struct_unit::hasItem(struct_item itm) { return itm.getOwner() == this; }

As stated, jass ops are much slower, and if possible those should be reduced to minimum.

wurst/lib/InventoryEvent.wurst
crojewsk: Issuing orders by string is slower than by ints.
And there is no complication, because in most cases you just use few of them and you are not creator of library. User should deal with exposed interface, not play with internal structures.

When playing with orders, you can have a order-id list open, such as: https://github.com/nestharus/JASS/blob/master/jass/Systems/OrderIds/script.j

wurst/lib/InventoryEvent.wurst
crojewsk: The behavior is changed in future update, I'll upload it today. I'd not use if-else if this wasn't void method.

wurst/lib/ItemRecipe.wurst
crojewsk: Precisely, in vjass version I have debug messages added. This is still not included in wurst version.

wurst/lib/ItemRestriction.wurst
crojewsk: Thanks!

wurst/lib/StackNSplit.wurst
crojewsk: Thanks!

wurst/lib/InventoryEvent.wurst
theQuazz: Makes sense, but to @jonathanmclaus' point can we do something like

```
if order >= ORDER_moveslot1 and order <= ORDER_useslot6
```

Once the constant-inliner runs it will become the same thing.

wurst/lib/ItemRecipe.wurst
```diff
@@ -0,0 +1,570 @@
...
+                destroy recipes
+                recipeMap.remove(itemTypeId)
+
+    /** Adds new entry to recipe ingredient list. */
+    function addItem(int itemTypeId, boolean remove, int charges)
+        LinkedList<ItemRecipe> recipes
+        int chrgs = charges
+
+        if itemTypeId <= 0 or itemTypeId == reward
+            return
```
theQuazz: Let's also show an error if someone tries to do this

wurst/lib/ItemRecipe.wurst
```diff
@@ -0,0 +1,570 @@
...
+        item itm
+        int charges
+        ItemData data
+        int idx
+        let recipeData = new HashMap<int, ItemData>()
+
+        let iter = ingredients.iterator()..next()
+        while iter.current != iter.dummy // Match() validated count > size already
+            itm = items.get(slot)
+            charges = GetItemCharges(itm)
```
theQuazz: Maybe we should put up a PR to add a `itm.getCharges()` method to the public library

wurst/lib/StackNSplit.wurst
```diff
@@ -0,0 +1,242 @@
...
+        if not isItemStackable(itemTypeId)
+            return false
+
+        return u.isInventoryFull() and not u.isItemFullyStacked(itemTypeId)
+
+init
+    registerPlayerUnitEvent(EVENT_PLAYER_UNIT_PICKUP_ITEM, () -> onPickUp())
+    registerInventoryEvent(EVENT_INVENTORY_MOVED, () -> onMoved())
+    registerSmoothItemPickupEvent(() -> onSmoothPickup())
+    addSmoothPickupCondition(new StackSmoothPickupPredicate())
```
theQuazz: We've made some customizations to stack and split which may need to be factored into this version some how. Specifically we made changes to how this library handles diseased meat.

https://github.com/island-troll-tribes/island-troll-tribes/blob/master/jass/systems/EasyItemStacknSplit.j#L139
https://github.com/island-troll-tribes/island-troll-tribes/blob/master/jass/systems/EasyItemStacknSplit.j#L484

Not sure if we have any other changes to the file (other than the stack table rather than item level)

wurst/lib/InventoryEvent.wurst
crojewsk: Nice! Didin't know wurst has the vjass equivalent there.

wurst/lib/InventoryEvent.wurst
```diff
@@ -0,0 +1,100 @@
...
+        for i = 0 to size - 1
+            if this.itemInSlot(i) == itm
+                return i
+    return -1
+
+function fireEvent(trigger evt, unit u, item itm, integer slotFrom, integer slotTo)
+    unit prevUnit = eventUnit
+    item prevItem = eventItem
+    int prevSlotFrom = eventSlotFrom
+    int prevSlotTo = eventSlotTo
```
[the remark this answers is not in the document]
crojewsk: This allows for nested events and code. Without those, one could experience unexpected behavior when handler registered for specific event causes another cascade of events to occur (fire event within an event). In other words, when cascade is done, no more event handlers are left to propagate event to, eventUnit could be set to different value than null.

This is true for any event propagation mechanism present in most languages.

wurst/lib/ItemRecipe.wurst
crojewsk: Debug messages are not part of current implementation. They will be added later. I don't want to add random message here and there, I want it to be organized - and this takes time.

wurst/lib/ItemRecipe.wurst
crojewsk: I'm in contact with Frotty quite often lately. I'll be moving few of these to wurstStdlib2 if allowed. I'm already a contributor to wurst 😀

jonathanmclaus:
Quazz should probably squash this to make sure it doesn't introduce any changes to version 3.0L before he publishes it.

PR #88, opened 2018-11-27
Updated -g, added -iamnoob, added wurst/_items folder

Jaccouille:
-g now give all gold when no amount is specified

added -iamnoob (still described as -i am noob in the quest log, didn't know how to update it)

added wurst/_items folder
added wurst/_items/GemOfKnowledge.wurst & removed it from the World Editor

commented a function in wurst/extensions/UnitExtensions.wurst which would produce a compile error due to a double definition, the other definition is in the dependency now


wurst/_items/GemOfKnowledge.wurst
```diff
@@ -0,0 +1,18 @@
+package GemOfKnowledge
+
+import ItemObjEditing
+
+@compiletime function create_gemofknowledge()
```
jonathanmclaus: createGemOfKnowledge

wurst/_items/GemOfKnowledge.wurst
```diff
@@ -0,0 +1,18 @@
...
+    new ItemDefinition('I03F', 'bspd')
+    ..setName("Gem of Knowledge")
+    ..setAbilities("A050,AIx1")
```
jonathanmclaus: Add a comment at the end of this line listing out the two names of these abilities.

wurst/_items/GemOfKnowledge.wurst
```diff
@@ -0,0 +1,18 @@
...
+    ..setTooltipBasic("Trade for Gem of Knowledge")
+    ..setHotkey("")
+    ..setTooltipExtended("Increase the growth rate of the wearer. Does not work for Beast Master, Hunter")
+    ..setDroppedWhenCarrierDies(true)
+    ..setLumberCost(40)
+    ..setGoldCost(0)
+    ..setInterfaceIcon("ReplaceableTextures\\CommandButtons\\BTNEnchantedGemstone.blp")
+    ..setDescription("Increases Growth Rate")
+    ..setLevel(0)
+    ..setCanBeSoldToMerchants(false)
```
jonathanmclaus: Do you know the default values for any of these? I think in general we'll want to write out some default values but others may just be noise. For example, if `setDroppedWhenCarrierDies` is true by default, we wouldn't want to write that out for every item.

wurst/systems/commands/Commands.wurst
```diff
@@ -87,3 +87,9 @@ init
             let msg = "Adjusted player color to {0}Player {1}|r".format(COLOR_CODE[id], (id + 1).toString())
             printTimedToPlayer(msg, 5, triggerPlayer)
             triggerPlayer.setColor(ConvertPlayerColor(id), true)
+            
+        registerCommandAll("iamnoob") (triggerPlayer, args) ->
```
jonathanmclaus: Changes this to `"recipes"` please.

wurst/_items/GemOfKnowledge.wurst
Jaccouille: I tested it with false value (just checking if it worked properly), still dropped when the carrier died

PR #92, opened 2018-11-29
3.2a

theQuazz:
- remove old school duel
- add -respawn-system
- fix docker setup for those OSX users (me)


wurst/systems/modes/Respawn.wurst
```diff
@@ -0,0 +1,61 @@
...
+import GameConfig
+import TimerUtils
+import TimerDialog
+
+@configurable constant GRACE_PERIOD_REVIVE_DELAY = 10.0
+
+function isGracePeriodActive() returns boolean
+    return udg_NOOB_TIME
+
+function quadraticReviveDelay(real x) returns real
```
jonathanmclaus: Change `x` to `seconds` or `num_seconds`. I'd prefer the latter but it's whatever.

theQuazz: We use `camelCase` not `snake_case` so I'll switch it to `numSeconds` and `numMinutes`

wurst/systems/modes/Respawn.wurst
```diff
@@ -1,16 +1,19 @@
...
 import TimerUtils
 import TimerDialog
+import ChatCommands
+import LinkedList
 
 @configurable constant GRACE_PERIOD_REVIVE_DELAY = 10.0
 
 function isGracePeriodActive() returns boolean
-    return udg_NOOB_TIME
+    return gameConfig.isGracePeriodEnabled() and udg_NOOB_TIME
```
jonathanmclaus: Are these not redundant?

theQuazz: No, one refers to the modes of the game and the other refers to whether noob time is still active

jonathanmclaus: Yes, but grace period is not enabled, isn't `udg_NOOB_TIME` always false?

theQuazz: Actually `udg_NOOB_TIME` is default `true` so hence the check

theQuazz:
@jonathanmclaus I've updated the PR to remove `udg_NOOB_TIME`

wurst/systems/modes/Respawn.wurst
```diff
@@ -80,6 +82,25 @@ init
...
                     delayTimer.doAfter(delay) ->
-                        revive(dyingUnit, owner)
                         timerDialog.display(false)
                         timerDialog.destr()
+
+                        let units = ENUM_GROUP
+                            ..enumUnitsOfForce(owner.getAllies())
+                            .toLinkedList()
+                            ..sortWith((a, b) -> begin
+                                let aType = a.isType(UNIT_TYPE_STRUCTURE) ? 1 : 0
```
jonathanmclaus: Could we cast a boolean into an integer here? If so, I think that's more readable.

wurst/systems/modes/Respawn.wurst
```diff
@@ -80,6 +82,25 @@ init
...
+                                let bType = a.isType(UNIT_TYPE_STRUCTURE) ? 1 : 0
+                                return aType == bType ?
```
jonathanmclaus: I'm not sure what your personal taste is, but I think it is generally recommended (and I prefer) to start both subsequent lines with the operator.

    return aType == bType
        ? a.getLevel() - b.getLevel()
        : aType - bType

theQuazz: I'm ok with that

PR #116, opened 2018-12-09
3.2a heat mana death

theQuazz:


wurst/systems/experience/Experience.wurst
```diff
@@ -32,11 +37,31 @@ init
     for i = 0 to PLAYER_NEUTRAL_PASSIVE - 1
         players[i].setHandicapXP(0)
 
+    doPeriodically(ANIMATION_PERIOD) (cb) ->
+        for troll in udg_trolls
+            if troll.getMana() < 1 and troll.isAlive()
+                let owner = troll.getOwner()
+                if owner.getGold() < 1
```
jonathanmclaus: Why does the dying troll need both 0 heat and mana? Shouldn't it only require one of the two?

theQuazz: Good call that would have been a bug. Fixed it

wurst/systems/experience/Experience.wurst
```diff
@@ -32,11 +36,30 @@ init
     for i = 0 to PLAYER_NEUTRAL_PASSIVE - 1
         players[i].setHandicapXP(0)
 
+    doPeriodically(ANIMATION_PERIOD) (cb) ->
+        for troll in udg_trolls
+            let owner = troll.getOwner()
+            if troll.getMana() < 1 or owner.getGold() < 1 and troll.isAlive()
```
jonathanmclaus: What are your thoughts on adding ITT-specific extentions for `troll.getEnergy` and `troll.getHeat`. This will allow the code to be more modular if we ever change these systems.

theQuazz: I don't mind `getHeat` but `getEnergy` seems overkill

jonathanmclaus: Could be. It'd be a pretty drastic change to necessitate that, so I'm fine without it.

PR #117, opened 2018-12-11
3.2a

theQuazz:


jonathanmclaus:
Why are there unmerged commits? Did someone push to 3.2a after this PR was closed?

theQuazz:
This was Paul adding commits which we weren’t actually going to include in 3.2a

Issue #118, opened 2018-12-11
Fix Hydra Fins


Awar2k:
Hydra Fins grant 22 armor.  Change to the value listed on the tooltip.

Awar2k:
Jaccouille is also saying that Hydra Fins have a spell shield.  Is this intended?

Awar2k:
Was an active ability added to Hydra Fins? Currently it is clickable but produces no effect when clicked.

PR #129, opened 2018-12-13
Fixes/hydraitems

Marsunpaisti:
Not sure how to do this since git doesnt understand .w3x files. **This .w3x has the fixes/mapfixes changes** in addition to hydraac fins being fixed to +4 armor and added spellshield in tooltip, removed agile feet ability which didnt work in the first place ( #118 ). 
Hydra claws also had the hydraac fins +18 armor ability without saying that in the description. Removed that aswell.


jonathanmclaus:
Unfortunately, GitHub doesn't handle binary files very well. This is part of why we're trying to transition to Wurst for object editing. Currently, just post in #developers that you are working on the map and hopefully no one else will work on it at the same time.

I would prefer renaming Hydraac Fins to Hydra Fins, as I don't think Hydraac is actually a word.

jonathanmclaus:
Does this duplicate any work in #128 ?

Marsunpaisti:
> 
> 
> Does this duplicate any work in #128 ?

To be honest I had no idea if the pull request also updates any commits I made after the request so I made a new one after making a small adjustment.

jonathanmclaus:
It does. You'll notice the commit logs for #127, #128, and #129 share
entries.

PR #131, opened 2018-12-14
Wurst stat loss

theQuazz:


wurst/systems/experience/StatLoss.wurst
```diff
@@ -0,0 +1,74 @@
...
+                )
+
+                for i = 0 to numStacksOfDepression
+                    energyLoss += GetRandomInt(1, 3)
+
+                if numStacksOfDepression > 1
+                    energyLoss += GetRandomInt(1, 6)
+                    if numStacksOfDepression > 2
+                        energyLoss += GetRandomInt(1, 6) + GetRandomInt(1, 6)
+
```
Marsunpaisti: Arent depression stacks taken into account twice here? Im not sure if its a mistake or if the original code was like this

theQuazz: I just did a 1:1 rewrite of the trigger. The old code would add 1-6 for each pair of depression buffs.

theQuazz:
@Marsunpaisti we can consider changing this system in the future, but for now I just wanted to do an exact rewrite

Issue #132, opened 2018-12-15
Smoked House meat bug


Jaccouille:
![](https://thumbs.gfycat.com/LazyHeavenlyArieltoucan-small.gif)

PR #137, opened 2018-12-16
Fixed meat smoking

Marsunpaisti:
Smokehouse didnt remove ingredients resulting in basically cloned meat https://github.com/island-troll-tribes/island-troll-tribes/issues/132
Now it removes them properly

cause of bug:
( unit.removeItemByID only drops items and doesnt delete it )


wurst/systems/spells/SmokeMeat.wurst
```diff
@@ -21,6 +20,13 @@ init
...
 
             if charges > 0
+                removeMeats(smoker)
                 let meat = createItem(ITEM_SMOKED_MEAT, smoker.getPos())
                 ..setCharges(charges)
-                UnitStackItem(smoker, meat)
\ No newline at end of file
+                UnitStackItem(smoker, meat)
+
+function removeMeats(unit smoker)
```
jonathanmclaus: Wouldn't this remove all meat that wasn't transformed?

Marsunpaisti: It does remove all meat, but all meat is transformed anyway in the function above.

jonathanmclaus: For some reason, I was thinking that only a limited amount of meat was transformed per casting.

PR #140, opened 2018-12-17
Mage spell reworks / recoded (Elementalist, dementia master)

Marsunpaisti:
New elementalist spells added to test mode. Give spellbook to any hero using -spells during -tm.

This one also (will) also contains a recoded dementia master dark gate that is currently available for preview with -spells2.


Marsunpaisti:
This should be basically playable for the dementia spell part. Elementalist spells are still behind testmode only.

PR #148, opened 2018-12-24
Added Hints to wurst, updated quest Log, fixed magic meteor

Jaccouille:
Tell me in which folder shall I move Hints.wurst


wurst/systems/Hints.wurst
```diff
@@ -0,0 +1,109 @@
...
+            Hint.addPLayer(p)
+
+    function display()
+        if IsPlayerInForce(GetLocalPlayer(), players)
+            printTimedToPlayer(GOLD_COLOR + "Hint: |r" + GENERAL_COLOR + message + "|r", 10, GetLocalPlayer())
+            if action != null
+                TriggerExecute(action)
+
+    static function initialize()
+        for i = 0 to 11
```
jonathanmclaus: This should be based off of a global I think.

wurst/systems/Hints.wurst
```diff
@@ -0,0 +1,109 @@
...
+    let hint = iterator.next()
+    hint.display()
+    if iterator.hasNext() == false
+        iterator.reset()
+        
+public function doPeriodicDisplay(LinkedList<Hint> hintList)
+        let iterator = hintList.iterator()
+        doPeriodically(TIME_BETWEEN_HINTS, (cb) -> displayNextHint(iterator))
+
+public function checkStat(unit troll)
```
jonathanmclaus: I don't think this belongs in the package for hints. Put it in the stat decay or its own package.

wurst/systems/Hints.wurst
```diff
@@ -0,0 +1,109 @@
...
+        this.action = CreateTrigger()
+        TriggerAddAction(this.action, what)
+
+    static function pingUnit(unit which)
+        pingLoc(GetUnitLoc(which))
+
+    static function pingLoc(location loc)
+        PingMinimapLocForForceEx(players, loc, 5, bj_MINIMAPPINGSTYLE_FLASHY, 100, 100, 100)
+
+    static function addPLayer(player p)
```
theQuazz: Typo on `addPLayer` (should be `addPlayer`)

wurst/systems/Hints.wurst
Jaccouille: I am not really ok with this suggestion, I put it in this package because I thought it would be convenient to use the hint instance to check the player in force in case they disabled the hints.

wurst/systems/Hints.wurst
Jaccouille: Can you elaborate on this? I don't fully understand what you mean/want me to do here.

theQuazz: @Jaccouille I'll give you an example--

```
for i = 0 to bj_MAX_PLAYERS - 1
    if players[i].isIngame()
        ...
```

wurst/systems/Hints.wurst
theQuazz: @Jaccouille I think its fine for now, we can always move it later. Perhaps you can periodically check every `gameConfig.getStatLossInterval()` and set the message duration to the same so that you'll only have one such message on your screen at a time (excluding the fade time). Might be a bit annoying but we can tweak it later.

wurst/systems/Hints.wurst
```diff
@@ -4,13 +4,13 @@ import LinkedList
 import ChatCommands
 import Game
 
-let TIME_BETWEEN_HINTS = 60.0
+let TIME_BETWEEN_HINTS = 2.0
```
theQuazz: Is this change intentional?

Jaccouille: woops

PR #156, opened 2018-12-26
Quests rewrite in Wurst

theQuazz:


Jaccouille:
Guess I'll have to update this one too

PR #157, opened 2018-12-26
Testversion 3.3 Beta 8 into master (updated)

Marsunpaisti:
Contains:
- New elementalist spells
- Fixed & recoded dementia master spells
- Integration of new elementalist into ID files, subbing into new ele, -oldele mode for playing with old elementalist
- Hatchery fix
- Smokehouse fix
- Mana potion / Health potion II - IV recipes removed fix
- Hydra items fix
- Steel coat fix
- New gatherer radars
- Telegather 100% chance for items (herbs only for herbmaster still)
- Item spawn rates balance
- Analysis.wurst
- Twin islands potion buff
- Metronome vision bug fix
- Added omni cure to sage
- Dementia fire summon 8 dps -> 10 dps
- Ice summon stun 0.35 sec -> 0.4 sec
- Fixed the glow tooltip
- Fixed Dementia fire summon debuff tooltip
- Added fire building + trapping demo/tutorial at mode selection


wurst/systems/spells/mage/dementia/DarkGateNew.wurst
```diff
@@ -0,0 +1,405 @@
...
+
+                if _summonTarget.value == null or _summonTarget.value.hasAbility('Bmlt') or _summonTarget.value.hasAbility('B00Q')
+                    if isValidAllyTarget(t)
+                        _summonTarget.value = t
+                
+            if _summonTarget.value != null
+                s.summonUnit.issueTargetOrderById(s.orderId, _summonTarget.value)
+                s.lastAttack = s.elapsedLifetime
+
+            destroy _summonTarget
```
theQuazz: Perhaps we can do something like

```
let pos = centerPos.toVec2()
forNearestUnit(pos, portalRadius, Filter(-> begin
    let filter = GetFilterUnit()
    return not filter.hasAbility('Bmlt') and not filter.hasAbility('B00Q') and isValidEnemyTarget(filter)
end) nearest ->
    if nearest != null
        s.summonUnit.issueTargetOrderById(s.orderId, nearest)
    else
        forNearestUnit(pos, portalRadius, Filter(-> isValidEnemyTarget(GetFilterUnit()))) nearest ->
            if nearest != null
                s.summonUnit.issueTargetOrderById(s.orderId, nearest)
```

Not a huge fan of the double `forNearestUnit` but it will work. Alternatively, we can do something like this:

```
import GroupExtensions

var targets = CreateGroup()
    ..enumUnitsInRange(pos, range, Filter(-> isValidTarget(GetFilterUnit()))
    .toLinkedList()

var without = targets.filter(u -> not u.hasAbility('Bmlt') and not u.hasAbility('B00Q'))
if without.size() > 0
    destroy targets
    targets = without

let target = targets
    ..sortBy((a, b) -> pos.distanceToSq(a.getPos()) - pos.distanceToSq(b.getPos()))
    .getFirst()

   s.summonUnit.issueTargetOrderById(s.orderId, target)

destroy targets
```

Added benefit of these options is that it provides an additional sort order on the units--their distance. If you dislike both these options, it is totally fine to stick with your current code. If you do, I would ask that you rename `unitObject` to `UnitObject` and I think you have duplicate code on lines 325-327. Happy to discuss further in Discord.

wurst/systems/spells/mage/dementia/DarkGateNew.wurst
```diff
@@ -0,0 +1,405 @@
...
+        var orb = this.orbs.get( GetRandomInt(0, orbs.size() - 1))
+        var centerPosVec2 = centerPos.toVec2()
+        var _foundTarget = false
+
+        //Find random target and cast + spawn effect on orb
+        forUnitsInRange(centerPosVec2, portalRadius) t ->
+            if not _foundTarget
+                if isValidEnemyTarget(t)
+                    orb.castRandomSpell(t)
+                    _foundTarget = true
```
theQuazz: I wonder if we should use something that may be *more* random like what I have below. `forUnitsInRange` may be deterministic and thus we'll always cast on the same unit.

```
let targets = CreateGroup..enumUnitsInRect(pos, range, Filter(-> isValidEnemyTarget(GetFilterUnit()))
let target = targets.getRandomUnit()
targets.destr()
```

wurst/systems/spells/mage/dementia/ActivateRunesNew.wurst
```diff
@@ -0,0 +1,120 @@
...
+                new DummyCaster(1).castTarget(owner, spell.spellId, 1, spell.orderId, target, r.pos)
+                destroy r
+
+function Runes.toString() returns string
+    if (this == Runes.Ka)
+        return "Ka"
+    else if (this == Runes.Nel)
+        return "Nel"
+    else
+        return "Lez"
```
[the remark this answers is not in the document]
Marsunpaisti: Added

wurst/systems/spells/mage/dementia/DementiaRunesNew.wurst
```diff
@@ -0,0 +1,178 @@
...
+            angle indexRuneAngle = runeAngleBase + offset * i.toReal()
+            r.setPos(ownerpos.offset2d(indexRuneAngle, ORBIT_DISTANCE))
+            r.setFacing(r.pos.angleTo2d(ownerUnit.getPos()))
+            i = i + 1
+
+    function addRune(DementiaRune rune)
+        if runes.size() < MAX_RUNE_COUNT
+            runes.add(rune)
+        else 
+            destroy runes.get(0)
```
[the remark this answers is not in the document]
Marsunpaisti: I created runes so that you only destroy them and dont worry about them later. The rune itself checks if its in the instances rune list and then removes itself in its onDestroy function.

wurst/systems/spells/mage/dementia/DarkGateNew.wurst
Marsunpaisti: The last option looks cleaner, the only thing is that the hasAbility checks are optional. So if there is a target with a buff we CAN attack that, but we prioritize targets without buffs. Can we do another sort where the other comparison function just returns 0 for any unit with no buffs and 1 unit if it has buffs?

wurst/systems/spells/mage/dementia/DarkGateNew.wurst
Marsunpaisti: Good point, in testing the current code SEEMED random so this never came up but that would ensure randomness.

wurst/systems/spells/mage/dementia/DarkGateNew.wurst
Marsunpaisti: added

PR #162, opened 2018-12-28
Cure all item cast wurst

theQuazz:


wurst/systems/spells/CureAllItem.wurst
```diff
@@ -0,0 +1,12 @@
...
+import InstantDummyCaster
+import RegisterEvents
+
+
+init
+    registerSpellEffectEvent(SPELL_CURE_ALL_ITEM) ->
+        let target = GetSpellTargetUnit()
+        let owner = target.getOwner()
+        let pos = target.getPos()
+        InstantDummyCaster.castTarget(owner, 'A02H', 1, "autodispel", target, pos)
```
jonathanmclaus: Why is casting twice needed?

theQuazz: Cures magic immunity then cures buffs. I'll try without it for now.

PR #163, opened 2018-12-28
Wurstify/units.zn

Marsunpaisti:


wurst/systems/core/init/Units.wurst
```diff
@@ -0,0 +1,33 @@
...
+
+public destructable MAMMOTH_GATE
+public unit MAMMOTH = null
+public unit MASTER_CRAFTER = null
+public unit OMINOUS_ALTAR = null
+public unit RECIPE_ARMORY = null
+public unit RECIPE_WORKSHOP = null
+public unit RECIPE_POT = null
+public unit RECIPE_WITCH_DOCTORS_HUT = null
+public unit RECIPE_TANNERY = null
```
theQuazz: These `RECIPE_*` can be removed because they aren't referenced anywhere else

wurst/systems/core/init/Units.wurst
```diff
@@ -0,0 +1,33 @@
+package Units
+import ClosureForGroups
+
+public destructable MAMMOTH_GATE
```
theQuazz: If you rename this variable (which I agree is a good idea) make sure that you update the references to it in Respawn.wurst and Game.wurst.

theQuazz: Or you can just remove it from this file and move the "finding" of this unit to Respawn.wurst (in that case I'd probably also add the initialization from Game.wurst for the gate to Respawn.wurst as well)

wurst/systems/core/init/Units.wurst
```diff
@@ -0,0 +1,33 @@
...
+public unit MAMMOTH = null
+public unit MASTER_CRAFTER = null
+public unit OMINOUS_ALTAR = null
```
theQuazz: Looks like we only have one reference to `OMINOUS_ALTAR`--I wonder if we should move this initialization to that location

https://github.com/island-troll-tribes/island-troll-tribes/blob/master/wurst/systems/core/Game.wurst#L38

theQuazz: Which we should probably move into https://github.com/island-troll-tribes/island-troll-tribes/blob/master/jass/crafting/Hydra%20Altar%20Buyingmake.j once we move it to Wurst

PR #167, opened 2018-12-29
Wurst replenish health

theQuazz:


wurst/systems/spells/priest/ReplenishHealth.wurst
```diff
@@ -0,0 +1,24 @@
...
+import RegisterEvents
+
+let SPELL_REPLENISH_HEALTH = 'A0F2'
+int CAST
+int ENDCAST
+
+function onCast()
+    let caster = GetTriggerUnit()
+    let unitTypeId = caster.getTypeId()
+    if unitTypeId != UNIT_MASTER_HEALER and unitTypeId != UNIT_MASTER_HEALER_1 and unitTypeId != UNIT_SAGE
```
jonathanmclaus: Why not use new entitities? This would break if skins were added.

jonathanmclaus: I'm not sure of the point of your first statement. Yes, some things break if skins are added. That doesn't mean new things need to. The entity class hierarchy is supposed to solve this.

PR #173, opened 2018-12-31
Jurstify

theQuazz:
There was a decent amount of work involved in converting Jass to Jurst, mostly adding a line like this to the top of our files. I don't know if it is worthwhile to read through this, I think testing it out is probably more efficient.

```
library LibraryName initializer onInit requires ID
```


wurst/systems/units/hostiles/trolls/Assassin.wurst
```diff
@@ -1,4 +1,5 @@
 package Assassin
+import ID
```
jonathanmclaus: Why did you add this to all entity classes? Seems like it isn't being used.

theQuazz: @jonathanmclaus it is being used. You probably have the old war3map.j.

PR #186, opened 2019-01-02
Fixed item stacking and bush spawns

Marsunpaisti:


wurst/lib/Stackables.wurst
```diff
@@ -2,6 +2,9 @@ package Stackables
 import StackNSplit
 import ID
 
+init
+    initStackables()
+    
 function initStackables()
     makeItemStackable(ITEM_ACID_BOMB, 2, 1)
```
jonathanmclaus: Can these not just be inferred? Is there not a runtime call to get the max charges of an item?

theQuazz: @jonathanmclaus even if there is we actually override it with these methods

jonathanmclaus: You're not overriding it if you're providing the same value. You're just introducing unnecessary magic numbers. Also, I don't believe it would be an override - you at least have to set it to a smaller value because the max charges of the item isn't being changed.

theQuazz: @jonathanmclaus in any case there isn't a way to get that property. People used to "hack" it by setting the item level to the max charges, but frankly I don't really like this approach. Also, that would require hiding a lot of our data in the binary, which I'm also opposed to.

jonathanmclaus: OK, good to just confirm that. Not too worried because this is trivial to sync once object definitions are in Wurst.

wurst/extensions/UnitExtensions.wurst
```diff
@@ -1,15 +1,15 @@
...
 import GameConfig
 import PublicLibrary
-
+import Brambles
 
 public let BUFF_ENSNARE_GROUND = 'Beng'
 public let BUFF_ENSNARE_AIR = 'Bena'
 public let BUFF_ENTANGLE = 'BEer'
-
+public let BUFF_ENTANGLE_BRAMBLES = getBramblesEntangleBuffId()
```
theQuazz: Generally we should prefer to do something like this:

```
public let BUFF_ENTANGLE_BRAMBLES = compiletime(getBramblesEntangleBuffId())
```

Marsunpaisti: I used a getter to give it a better public name since the brambles packages own variable name is DUMMY_ENTANGLE_BUFF and using that name would be misleading, eventually when we we get to the point where we have many of these buffs being added

theQuazz: @Marsunpaisti maybe we can use a more explicit name in its own package--like `BUFF_ENTANGLE_BRAMBLES` like you've used here

PR #201, opened 2019-01-04
Fix unit entity

theQuazz:


wurst/systems/units/UnitEntity.wurst
```diff
@@ -32,8 +32,8 @@ public abstract class UnitEntity
     static function register(int unitTypeId, Function<unit, thistype> factory)
         factories.put(unitTypeId, factory)
 
-    static function findForUnit(unit hostile) returns thistype
-        return instances[hostile.getIndex()]
+    static function findForUnit(unit whichUnit) returns thistype
```
jonathanmclaus: Why not just `u`? This comes up enough that a standardized name would be good.

wurst/systems/units/hostiles/Hostile.wurst
```diff
@@ -36,9 +37,6 @@ public abstract class Hostile extends UnitEntity
         for i = 1 to numCorpses
             createCorpse(players[PLAYER_NEUTRAL_AGGRESSIVE], UNIT_MEAT, pos, randomAngle())
 
-        let iter = drops.iterator()
-        for drop from iter
-            createItem(drop, pos)
-        iter.close()
+        drops.each(drop -> createItem(drop, pos))
```
jonathanmclaus: Do you need to destroy `drops` after this?

wurst/systems/units/UnitEntity.wurst
theQuazz: `whichUnit` is also standard, that is what they use in common.j and throughout the stdlib

wurst/systems/units/hostiles/Hostile.wurst
theQuazz: `.each` will cleanup the list unless it is owned (which this list is not)

PR #213, opened 2019-01-05
UnitEntity optimization

Marsunpaisti:
UnitEntities now generate drops list on death instead of on creation

### DIFF
--- a/wurst/systems/units/hostiles/AncientHydra.wurst
+++ b/wurst/systems/units/hostiles/AncientHydra.wurst
@@ -32,8 +32,8 @@ public class AncientHydra extends Hostile
     override function postCreate()
         super.postCreate()
         let hydra = getUnit()
-        ..addSleep(false)
-        ..setCreepGuard(false)
+            ..addSleep(false)
+            ..setCreepGuard(false)
 
         HornSound()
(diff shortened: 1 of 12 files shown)


wurst/systems/units/hostiles/RearingAnimal.wurst
```diff
@@ -13,7 +14,11 @@ public abstract class RearingAnimal extends BonyAnimal
 
     override function postCreate()
         super.postCreate()
-        drops.add(getHideTypeId())
+
+    override function getDrops() returns LinkedList<int>
+        let drops = super.getDrops()
+        ..add(getHideTypeId())
```
jonathanmclaus: Indent.

wurst/systems/units/hostiles/AncientHydra.wurst
```diff
@@ -73,6 +73,8 @@ public class AncientHydra extends Hostile
         for i = 0 to 1
             new LesserHydra(pos)
 
+        super.onDeath()
```
Marsunpaisti: So was this supposed to be here? Quazz at first said no but now its still here

theQuazz: @Marsunpaisti I actually think it is supposed to be there now

PR #227, opened 2019-01-08
Changed unit entity tracker to use map instead of array to avoid collisions

Marsunpaisti:
Unit entities stored in a hashmap instead of in an array by unitindexes, fixed error msg on lesser hydra death. Also .w3x changes for hydra items.


wurst/systems/units/UnitEntity.wurst
```diff
@@ -50,7 +53,9 @@ public abstract class UnitEntity
         postCreate()
 
     function postCreate()
-        instances[theUnit.getIndex()] = this
+        if instances.has(theUnit)
```
jonathanmclaus: This is weirdly formatted.

wurst/systems/units/UnitEntity.wurst
```diff
@@ -11,7 +11,7 @@ public function randomAngle() returns angle
 
 
 public abstract class UnitEntity
-    static thistype array instances
+    static HashMap<unit, thistype> instances = new HashMap<unit,thistype>()
```
jonathanmclaus: What is the overhead on this?

wurst/systems/units/UnitEntity.wurst
Marsunpaisti: Leftovers from debug code

wurst/systems/units/UnitEntity.wurst
Marsunpaisti: No idea to be honest but I'd like to try it to see if its more reliable than using unitindex

wurst/systems/units/hostiles/LesserHydra.wurst
```diff
@@ -95,9 +94,8 @@ public class LesserHydra extends Hostile
 
     override function onDeath()
         deathPos = getPos()
-        removeUnit()
```
theQuazz: We definitely want to remove the unit here

Marsunpaisti: Are you sure about that? The ondestroy function removes the unit later

PR #239, opened 2019-01-11
Fixed Depression to reduce movespeed

Marsunpaisti:
Now properly slows targets, is a single-target ability, and improved tooltips for the ability and buffs.


wurst/systems/spells/mage/Depress.wurst
```diff
@@ -1,15 +1,50 @@
...
 function onCast()
-    var target = GetSpellTargetUnit()
-    if target.getMana() < 40
-        return
-    target.subMana(10)
-    doPeriodicallyCounted(2.00, 4, (cb) -> target.subMana(5))
+    let target = GetSpellTargetUnit()
+    let caster = GetSpellAbilityUnit()
+    let owner = caster.getOwner()
+    doAfter(0.2) ->
```
jonathanmclaus: why did you add the delay?

Marsunpaisti: To approximate projectile flight time. Without a custom projectile I dont see a way to detect the impact otherwise.

Marsunpaisti: That doesnt differentiate between the exact ability that deals the damage though. And also a lot of effort for just a simple effect like this to have (even more) correct timing

theQuazz: You’d have to cast it from a dummy then detect if the dummy was the damage source

theQuazz: I agree it’s unnecessary though

PR #244, opened 2019-01-12
Wurstified Mage Masher

Marsunpaisti:


wurst/_items/MageMasher.wurst
```diff
@@ -0,0 +1,58 @@
...
+import IttDummyCaster
+import Orders
+import StandardTextTags
+import LinkedList
+
+let MANABURN_ID = compiletime(ABIL_ID_GEN.next())
+let DRAIN_AMOUNT = 6.
+
+LinkedList<int> MASHER_TARGET_CLASSES = asList(
+                            UNIT_MAGE,
```
jonathanmclaus: Do not hard code lists like this. It's not resilient to more skins. Use `instanceof` instead.

wurst/_items/MageMasher.wurst
```diff
@@ -0,0 +1,58 @@
...
+    if isValidMasherTarget(owner, target)
+        new IttDummyCaster(0.45).castTarget(owner, MANABURN_ID, 1, Orders.manaburn, target, attacker.getPos())
+
+function isValidMasherTarget(player attackOwner, unit target) returns bool
+    return target.getOwner().isEnemyOf(attackOwner)
+        and target.isAlive()
+        and MASHER_TARGET_CLASSES.has(target.getTypeId())
+
+init
+    EventListener.add(EVENT_UNIT_DAMAGED) ->
```
jonathanmclaus: Is there not a way to use an `EVENT_UNIT_DAMAGE` event for detecting when abilities like depression land?

Marsunpaisti: Yes, but for depres that was deemed unnecessary and quite a bit of work just to apply the slow with correct timing.

PR #251, opened 2019-01-16
Changed baby animals to move off of drops upon spawn

Marsunpaisti:
Baby animals move slightly on spawn so they wont interfere with item pickups

### DIFF
--- a/wurst/systems/units/hostiles/BabyAnimal.wurst
+++ b/wurst/systems/units/hostiles/BabyAnimal.wurst
@@ -1,0 +1,21 @@
+package BabyAnimal
+import ID
+import BonyAnimal
+import Orders
+import GeometryUtils
+
+let SPAWN_WANDER_DISTANCE = 60.
+
+public abstract class BabyAnimal extends BonyAnimal
+    construct(unit whichUnit)
+        super(whichUnit)
+
+    construct(player whichPlayer, vec2 pos)
+        super(whichPlayer, pos)
+
+    construct(player whichPlayer, vec2 pos, angle facing)
+        super(whichPlayer, pos, facing)
+
+    override function postCreate()
+        super.postCreate()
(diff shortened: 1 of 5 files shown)


wurst/systems/units/hostiles/BabyAnimal.wurst
```diff
@@ -0,0 +1,25 @@
...
+    construct(player whichPlayer, vec2 pos)
+        super(whichPlayer, pos)
+
+    construct(player whichPlayer, vec2 pos, angle facing)
+        super(whichPlayer, pos, facing)
+
+    override function postCreate()
+        super.postCreate()
+        getUnit().issuePointOrderById(Orders.move, getPos().polarOffset(randomAngle(), SPAWN_WANDER_DISTANCE))
+        getUnit().addAbility(GHOST_VIS_ID) //Add ghost (visible)
```
jonathanmclaus: I think comment is redundant. Where is the ID package coming from?

Marsunpaisti: Built-in to wurst. Comment removed

jonathanmclaus: So that isn't actually in the ID package, which is where I thought it was. Do you know what the ID package is for?

I would add the visible Ghost ability [here](https://github.com/wurstscript/WurstStdlib2/blob/200f171296d017d262aec8ad67b3962cdedd4e67/wurst/_wurst/assets/AbilityIds.wurst#L128).

Where you're getting it now seems to be [here](https://github.com/wurstscript/WurstStdlib2/blob/0aa2173acff1ba53fa56d70821c9e86fdcfb0548/wurst/_wurst/Basics.wurst#L21). But the `Basics` package isn't imported so I'm not sure how you're using it.

PR #252, opened 2019-01-16
Imported troll cursor

Marsunpaisti:


theQuazz:
No, we’re not doing this

jonathanmclaus:
This doesn't change any UI by itself - it just imports the file. I thought there was another race we're changing, so I'm approving this but not the actual UI change yet.

PR #254, opened 2019-01-16
Move trollcursor to undead

Marsunpaisti:


jonathanmclaus:
I'm going to accept this even though I don't think we should overwrite. I'll just put an issue to track that change.

PR #258, opened 2019-01-17
Removed boat collision

Marsunpaisti:


wurst/systems/trade/TradeShipSupervisor.wurst
```diff
@@ -34,9 +34,11 @@ public class TradeShipSupervisor
         let enteringUnit = GetEnteringUnit()
         if ships.has(enteringUnit)
             ships.get(enteringUnit).processEnterRegion()
-            
+
     construct(TradeShipStop firstStop, int shipId)
         ship = createUnit(players[PLAYER_NEUTRAL_PASSIVE], shipId, firstStop.getPos(), angle(0))
+        ship.setPathing(false)
+        ship.addAbility(GHOST_VIS_ID)
```
theQuazz: Does it need both of these?

Marsunpaisti: I didnt test it but since .setPathing alone doesnt change the pathfinding to ignore terrain I assumed other units try to avoid the no-collision unit as well unless it has ghost.

theQuazz: Ok. Did you test that the boats don't go through land?

PR #263, opened 2019-01-18
Fixed error where spell damage could activate Mage Masher

Marsunpaisti:


wurst/_items/Blowgun.wurst
```diff
@@ -136,6 +136,9 @@ class Blowpipe
             simError(caster.getOwner(), "No ammo!")
             return
         else
+            //Flat 10 damage + ammo specific effect
+            UnitDamageTarget(caster, this.target, 10, false, false, ATTACK_TYPE_NORMAL, DAMAGE_TYPE_UNIVERSAL, null)
```
jonathanmclaus: Please use `caster.damageTarget(this.target, 10, ATTACK_TYPE_NORMAL)`, and that method signature for all other calls in this diff.

Marsunpaisti: Updated

PR #272, opened 2019-01-22
Added more Lodash functionality

theQuazz:


wurst/lib/Lodash.wurst
```diff
@@ -305,6 +315,199 @@ public function each<T>(VoidFunction<T> func, LinkedList<T> list)
...
 
+public function zip<A, B>(LinkedList<A> a, LinkedList<B> b) returns IterableMap<A, B>
+    let map = new IterableMap<A, B>()
+    let iterA = a.iterator()
+    let iterB = b.iterator()
+    while iterA.hasNext()
+        let bVal = iterB.hasNext()
+            ? iterB.next()
+            : null
+        map.put(iterA.next(), bVal)
```
jonathanmclaus: Is this how `zip` works in Lodash? I haven't used it before but the Python version ends with the shorter iterable, rather than automatically padding it with nulls. Also seems weird to have the operation not be commutative.

wurst/lib/Lodash.wurst
```diff
@@ -305,6 +315,199 @@ public function each<T>(VoidFunction<T> func, LinkedList<T> list)
...
+        curr += incr
+        return ret
+
+    function close()
+        destroy this
+
+    function iterator() returns Range
+        return new Range(min, max, incr)
+
+    function toList() returns LinkedList<int>
```
jonathanmclaus: Is there not general functionality for converting an iterator to a list?

wurst/lib/Lodash.wurst
theQuazz: Indeed that is how `zipObject` works in Lodash. I can change it though.

wurst/lib/Lodash.wurst
theQuazz: @jonathanmclaus not that I'm aware of

wurst/lib/Lodash.wurst
theQuazz: @jonathanmclaus yeah I guess I've implemented `zipObject`. I'll rename and add `zip`.

jonathanmclaus: Yeah, I'm an idiot and was going off the name. I think it's fine to map to nulls in that case.

theQuazz: @jonathanmclaus updated

wurst/lib/Lodash.wurst
```diff
@@ -305,6 +316,341 @@ public function each<T>(VoidFunction<T> func, LinkedList<T> list)
...
+    a.maybeFree()
+    b.maybeFree()
+    destroy stored
+    return lst
+
+public function LinkedList<T>.intersection<T>(LinkedList<T> lst) returns LinkedList<T>
+    return intersection(this, lst)
+
+public function difference<T>(LinkedList<T> a, LinkedList<T> b) returns LinkedList<T>
+    T array occurences
```
jonathanmclaus: Why do you use an array here instead of a HashSet?

wurst/lib/Lodash.wurst
```diff
@@ -305,6 +316,341 @@ public function each<T>(VoidFunction<T> func, LinkedList<T> list)
...
+
+    construct(int start, int finish, int incr)
+        this.start = start
+        this.finish = finish
+        this.incr = incr
+        curr = start
+
+    function hasNext() returns bool
+        return incr > 0 ? curr < finish : curr > finish
+
```
Marsunpaisti: What if the increment will be large enough that the next call would be outside the upper bound?

wurst/lib/Lodash.wurst
```diff
@@ -305,6 +316,341 @@ public function each<T>(VoidFunction<T> func, LinkedList<T> list)
...
+    function next() returns int
+        let ret = curr
+        curr += incr
+        return ret
+
```
Marsunpaisti: is hasNext() supposed to be called manually before using next()? Looks like this can go over the upper bound if being called without hasNext()

theQuazz: @Marsunpaisti `for` loops always call `hasNext()` before `next()`

wurst/lib/Lodash.wurst
theQuazz: It may happen in the case of integer overflow but otherwise it likely wouldn't happen because `hasNext` is always called before `next`

wurst/lib/Lodash.wurst
theQuazz: @jonathanmclaus good point, I've switched it to a HashSet. Array is more performant though so I might switch it back for most types and use the `HashSet` for primitive types (other than int and maybe bool) and handles.

jonathanmclaus: Are you going to do that now or should I accept? Not really clear on the difference between the built-in `array` and `initHashtable`.

wurst/lib/Lodash.wurst
```diff
@@ -17,71 +17,53 @@ public abstract class Callable
         if cleanup
             destroy this
 
-    function destroyNextTick()
-        nullTimer(-> destroy this)
-
+public abstract class EmptyVoidFunction extends Callable
```
jonathanmclaus: Where is this used? You added it but I don't see anything that references it.

wurst/lib/Lodash.wurst
theQuazz: Should be updated

wurst/lib/Lodash.wurst
theQuazz: Hmm I may not have ended up using it. Either way it could be useful in the future.

wurst/lib/Lodash.wurst
theQuazz: Ah I meant I'll leave it for a later PR

PR #287, opened 2019-01-29
Fix -all gatherer (and other classes)

theQuazz:


theQuazz:
Class is a reserved word. And I don’t think that todo is necessary because
beast master works fine.

wurst/systems/modes/GameMode.wurst
```diff
@@ -10,6 +10,7 @@ import TimerUtils
 import LegacyColors
 import IDUtils
 import ChatUtils
+import Lodash
```
jonathanmclaus: Adding this caused a call to `filter` to be ambiguous.

PR #288, opened 2019-01-29
Show item charges while on ground

Marsunpaisti:
Appends "(x itm.getCharges())" on items when dropped, and removes the suffix on pickup

### DIFF
--- a/wurst/systems/core/ItemAmount.wurst
+++ b/wurst/systems/core/ItemAmount.wurst
@@ -1,0 +1,29 @@
+package ItemAmount
+
+import ClosureEvents
+import StackNSplit
+
+function onDrop(item itm)
+    if isItemStackable(itm.getTypeId())
+        var name = itm.getName()
+        int foundIndex = name.lastIndexOf(" (x")
+        if foundIndex != -1
+            name = name.substring(0, foundIndex)
+
+        name = name + " (x" + itm.getCharges().toString() + ")"
+        itm.setName(name)
+
+function onPickup(item itm)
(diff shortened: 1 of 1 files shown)


wurst/systems/core/ItemAmount.wurst
```diff
@@ -0,0 +1,29 @@
+package ItemAmount
+
+import ClosureEvents
+import StackNSplit
+
+function onDrop(item itm)
+    if isItemStackable(itm.getTypeId())
+        var name = itm.getName()
+        int foundIndex = name.lastIndexOf(" (x")
```
jonathanmclaus: Abstract lines 9 - 11 into a function. You're doing the exact same thing in the `onPickup`.

Marsunpaisti: updated

wurst/systems/core/ItemAmount.wurst
```diff
@@ -0,0 +1,28 @@
...
+function getItemBaseName(item itm) returns string
+    let name = itm.getName()
+    int foundIndex = name.lastIndexOf(" (x")
+    if foundIndex != -1
+        return name.substring(0, foundIndex)
+
+    return name
+
+init
+    registerPlayerUnitEvent(EVENT_PLAYER_UNIT_PICKUP_ITEM) ->
```
jonathanmclaus: Are these two events enough to cover all cases, like unit death and quickdrop?

Marsunpaisti: Unit.RemoveItem causes drop events so yes

PR #297, opened 2019-02-05
Added tooltips describing difficulty of play during class selection

Marsunpaisti:


jonathanmclaus:
This changes a few code files because new objects were copied from the Shapeshifter version of the .w3x into the version published by Jaccouille, which causes object IDs to change.

PR #299, opened 2019-02-05
Blowgun ammo contents tooltip suffix removed

Marsunpaisti:
Now that the itemcount system changes item names to for example, "Thistles (x3)" the blowgun was showing this twice such as "Thistles (x3) x3". Now the blowgun no longer adds a suffix of its own when listing ammo contained.

### DIFF
--- a/wurst/_items/Blowgun.wurst
+++ b/wurst/_items/Blowgun.wurst
@@ -219,7 +219,7 @@ class Blowpipe
                 var charges = itm.getCharges()
                 if (charges == 0)
                     charges = 1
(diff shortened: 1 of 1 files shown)


jonathanmclaus:
How does this work when containing multiple ammo types? I'm guessing that 4x dark thistles + 7x regular thistles would result in the in-inventory tooltip discluding amounts and the mouse-over name when it is on the ground would simply say x11. Is that correct?

Marsunpaisti:
Blowgun itself is not a stackable so it wont get the charges-suffix. And for items inside the blowgun with 4x dark thistles and 7x regular it would show
Contained ammunition:
Dark thistle (x4)
Thistles (x7)

PR #300, opened 2019-02-05
Pump up and anti-magic shell changes

Marsunpaisti:
**(PRIEST)** Anti-Magic shell is now AoE for every ally in 800 radius. 20 second cooldown with a 10 second duration.
Anti-Magic shell **potion** has the same cooldown as before, duration 20s -> 30s

Pump up removed from priest first level spellbook
Pump up cooldown from 30s -> 40s, Attack speed increase from 30% -> 20%


Marsunpaisti:
I think at least liquid and mambo were cool with this

PR #302, opened 2019-02-05
Scroll of cyclone allows walking through cycloned unit

Marsunpaisti:


wurst/_items/ScrollOfCyclone.wurst
```diff
@@ -0,0 +1,20 @@
+package ScrollOfCyclone
+
+import ClosureEvents
+import ClosureTimers
+
+let ABILITY_SCROLL_OF_CYCLONE = 'S000'
+let BUFF_CYCLONE = 'Bcyc'
```
jonathanmclaus: Do buff IDs normally have lowercase? Seems odd that it's all uppercase most of the time but not here.

wurst/_items/ScrollOfCyclone.wurst
```diff
@@ -0,0 +1,20 @@
...
+let BUFF_CYCLONE_EXTRA = 'Bcy2'
+
+function onCyclone()
+    let target = GetSpellTargetUnit()
+    target.addAbility(GHOST_VIS_ID)
+    doPeriodicallyTimed(0.1, 15) buffPoller ->
```
jonathanmclaus: You can't end cyclone early such as using a cure. So you could just use `doAfter`.

wurst/_items/ScrollOfCyclone.wurst
```diff
@@ -0,0 +1,20 @@
+package ScrollOfCyclone
+
+import ClosureEvents
+import ClosureTimers
+
+let ABILITY_SCROLL_OF_CYCLONE = 'S000'
+let BUFF_CYCLONE = 'Bcyc'
+let BUFF_CYCLONE_EXTRA = 'Bcy2'
```
jonathanmclaus: Isn't this an ability ID? And what is it used for?

PR #307, opened 2019-02-07
Allowed allies to transfer from bushes via selection.

jonathanmclaus:


Marsunpaisti:
Whats the endgame of this? Just to allow allies to click the bush for the scout/thief?

jonathanmclaus:
@Marsunpaisti Yes, exactly like the title says. This is mainly helpful when controlling units of leavers.

Marsunpaisti:
Ah, I see

wurst/systems/spawns/BushSelection.wurst
```diff
@@ -1,23 +1,43 @@
...
+
+    // Enumerate all units that may be close enough to receive items.
+    GroupEnumUnitsInRange(
+        ENUM_GROUP,
+        u.getPos().x,
+        u.getPos().y,
+        MAX_DIST.toReal(),
+        Filter(-> true)
+    )
+
```
Marsunpaisti: Im not sure if you consider this cleaner but
```
let units = CreateGroup()..
    enumUnitsInRange(pos, radius, null)
```

jonathanmclaus: I wasn't aware of that method. I'll use it when writing up `findNearestUnit`.

wurst/utils/UnitUtils.wurst
```diff
@@ -1,12 +1,14 @@
 package UnitUtils
+import Lodash
 
-public function findNearestUnit(vec2 pos, real range, filterfunc filter) returns unit
```
theQuazz: I would keep both versions of this function (let's allow an overload). In fact, to make the interface better, lets add this as well:

```
public function findNearestUnit(vec2 pos, real range, code filter) returns unit
    let filterFunc = Filter(filter)
    GroupEnumUnitsInRange(ENUM_GROUP, pos.x, pos.y, range, filterFunc)
    unit nearest = null
    var bestDist = REAL_MAX
    for u from ENUM_GROUP
        let distSq = pos.distanceToSq(u.getPos())
        if distSq < bestDist and predicate.call(u)
            nearest = u
            bestDist = distSq
    predicate.maybeFree()
    filterFunc.destr()
    return nearest
```

wurst/systems/units/hostiles/Fawn.wurst
```diff
@@ -63,7 +63,7 @@ public class Fawn extends BabyAnimal
             fawn.issuePointOrderById(Orders.move, acornPos)
             return
 
-        let troll = findNearestUnit(acornPos, 500, Filter(-> GetFilterUnit().isTroll()))
+        let troll = findNearestUnit(acornPos, 500, target -> target.isTroll())
```
theQuazz: Once you add the other variants back we can do this instead:
```
let troll = findNearestUnit(acornPos, 500, -> GetFilterUnit().isTroll())
```

wurst/systems/spawns/BushSelection.wurst
```diff
@@ -1,23 +1,34 @@
...
+            and id == getTrollBaseClass(target)
+
+    // Deliver the items
+    if match != null
+        u.transferItems(match)
+
+
+init
+    registerPlayerUnitEvent(EVENT_PLAYER_UNIT_SELECTED) ->
+        transferItems(GetTriggerUnit(), GetTriggerPlayer())
```
theQuazz: This looks good 👍

wurst/utils/UnitUtils.wurst
jonathanmclaus: What's the difference between a `code` and a `filterfunc`?

theQuazz: `code` gets converted to `filterfunc` by passing it to `Filter`

jonathanmclaus: I got that from the code. I mean what are the two for? `code` just looks an arbitrary `-> boolean` lambda. Is `Filter` just some arbitrary construct used for `GroupEnumUnitsInRange`?

theQuazz: Neither are a lambda, they are both basic types that JASS supports. `code` can return `boolean` or `nothing` where as `Filter` should only take `code` variables which return `boolean`. `Filter` is used in other places as well, but it does seem like a somewhat arbitrary construct. There is more info here: http://jass.sourceforge.net/doc/types.shtml

PR #309, opened 2019-02-07
Update brambles to work with new stdlib2

Marsunpaisti:


Marsunpaisti:
This is included in wurst spears pr

jonathanmclaus:
Why not submit this separately? It's a separate feature that could have its own PR.

Marsunpaisti:
I think the spear branch was just about to be merged. That is also basically the branch that caused the need for the stdlib change.

PR #314, opened 2019-02-10
Updated -ms command, now give ms of the selected unit

Jaccouille:
Awar triggered me with a movement speed conversation, so I add to check it out ingame, realized -ms doesn't work very well in -dev mode.
I think this would be better :
mscmdupdate

### DIFF
--- a/wurst/systems/commands/Commands.wurst
+++ b/wurst/systems/commands/Commands.wurst
@@ -3,6 +3,7 @@ import LinkedList
 import ChatCommands
 import StringExtensions
 import PlayerExtensions
+import UnitExtensions
 import ChatUtils
 import Game
 import GameMode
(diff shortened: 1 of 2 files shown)


wurst/systems/commands/Commands.wurst
```diff
@@ -95,11 +96,19 @@ init
...
 
         registerCommandAll("ms") (triggerPlayer, args) ->
-            let hero = triggerPlayer.getTroll()
-            if hero.isAlive()
-                let msg = "Your movement speed is {0}".format(hero.getMoveSpeed().toString().color(ENERGY_COLOR))
+            let units = CreateGroup()..enumUnitsSelected(triggerPlayer, null)
+            if units.isEmpty()
+                let hero = triggerPlayer.getTroll()
+                if hero.isAlive()
+                let msg = "{0} the {1} movement speed is {2}".format(hero.getProperName().color(HIGHLIGHT_COLOR), hero.getName().color(SPECIAL_COLOR), hero.getMoveSpeed().toString().color(HIGHLIGHT_COLOR))
```
jonathanmclaus: This should be indented I think.

PR #321, opened 2019-03-01
Ensnared hostile will now abort the move order

Jaccouille:
From what I saw, creep behavior seems to issue an move order back to initial position after ~6 seconds.
Made small bunch of code which will abort the move order when the unit's being targeted by ensnare ability.
showcase

Also added a "-hostile" command to give selected unit to PLAYER_NEUTRAL_AGGRESSIVE

### DIFF
--- a/wurst/systems/commands/SpawningTestHelpers.wurst
+++ b/wurst/systems/commands/SpawningTestHelpers.wurst
@@ -46,3 +46,8 @@ init
                                 break
                     else
                         simError(triggerPlayer, "Cannot spawn {0}".format(idStr))
+
+            registerCommandAll("hostile") (triggerPlayer, args) ->
+                let units = CreateGroup()..enumUnitsSelected(triggerPlayer, null)
+                for u in units
+                    u.setOwner(players[PLAYER_NEUTRAL_AGGRESSIVE], true)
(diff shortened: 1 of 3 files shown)


wurst/systems/commands/SpawningTestHelpers.wurst
```diff
@@ -46,3 +46,8 @@ init
                                 break
                     else
                         simError(triggerPlayer, "Cannot spawn {0}".format(idStr))
+
+            registerCommandAll("hostile") (triggerPlayer, args) ->
+                let units = CreateGroup()..enumUnitsSelected(triggerPlayer, null)
+                for u in units
+                    u.setOwner(players[PLAYER_NEUTRAL_AGGRESSIVE], true)
```
theQuazz: You need to also call `units.destr()` at the end and/or call `for u from units`

wurst/systems/spawns/HostileNetBehavior.wurst
```diff
@@ -0,0 +1,27 @@
+package HostileNetBehavior
+import ClosureTimers
+import RegisterEvents
+import ID
+
+constant let MOVE_ORDER = String2OrderIdBJ("move")
```
theQuazz: You could also use `Orders.smart` from package `Orders`

Jaccouille: /** This is a point or object targeted order that is like a right click. */

I don't really get it, what's the purpose?

theQuazz: Or just `Orders.move`

Jaccouille: Oh yeah I could use that

PR #323, opened 2019-03-04
Shield update from snap suggestion

Jaccouille:
I scrolled up the balance channel and find snap suggestion on shield with 10 thumbs up, I kinda liked it so I made it.
Shield have 100% chance to reduce damage by 4/6/10/13, doesn't give strenght, add 0/1/2/4 armor, 10% spell resistance.
Steel Shield has a bash ability who silences and disarms target for 2 seconds, has 15 seconds cooldown
ShowCase

I also extended the demo on the recipe islands so I can ping it when new player have trouble with food/killing elk

I had to create ability object for bonus armor because I didn't manage to use the basic one, it just didn't work, could use some help if you know how to make it work, thx.


Jaccouille:
Merged in #334

PR #326, opened 2019-03-14
Troll unit Factory

Jaccouille:
Just made a factory to generate Troll UnitDefinition, alongside the tooltips.

Update : I think that's a decent beginning that can be improved later on. Still need to make the shapeShifter


Jaccouille:
Merged in #334

PR #331, opened 2019-03-18
WurstSetup arg have been updated

Jaccouille:
url


Jaccouille:
https://github.com/island-troll-tribes/island-troll-tribes/pull/335

PR #332, opened 2019-03-19
Wurstified Troll Upgrade

Jaccouille:
Ok so I was working on the TrollUnitGenerator branch to test out those new upgrade, seemed to work correctly to me, prolly still got some bugs, have only 1 in mind for now.
This is supposed to work with the TrollUnitGenerator branch, won't work without it.
I cherry picked those files from TrollUnitGenerator branch, trying to split up the pr, gonna describe quickly.

1. I disabled troll upgrade & medal in jurst.
2. Added a bunch of ID from the GUI so I could use them for the TrollUnitGenerator (mostly spells).
3. Wurstified medallion of courage system.
4. Added TrollUnitTextConstant which is used for transform spell description here.
5. Added IdListConstant which I use here to check if unit currently using medal is subbed or not
6. Added & Wrapped up transform spell definition

Big issue right now : this TrollUpgrade I wurstified got one problem, it doesn't transfer item to the newUnit, I believe that's because I added the Inventory ability through the troll classes (gatherer.wurst, not added in this PR), so when I create the newUnit, it doesn't have inventory yet and the item end up on ground.
I can add them at the compiletime Definition using my brand new generator, it works, but it's ugly to look at and sound dumb to me.


wurst/_items/MedallionOfCourage.wurst
```diff
@@ -0,0 +1,93 @@
...
+    let caster = GetSpellAbilityUnit()
+    let owner = caster.getOwner()
+    let unitId = caster.getTypeId()
+    var _superSubUpgradeId = 0
+
+    if unitId == UNIT_HUNTER or unitId == UNIT_BEAST_MASTER
+        simError(owner, "This unit cannot use this item, read the item description")
+        return
+    if  SUB_TROLL_UNIT_LIST.has(unitId) or SUPERSUB_TROLL_UNIT_LIST.has(unitId)
+        simError(owner, "You need to be a basic troll (mage, priest, thief, scout, gatherer) in order to use this item")
```
theQuazz: Missed a return

wurst/_items/MedallionOfCourage.wurst
```diff
@@ -0,0 +1,93 @@
...
+            caster.addAbility(_superSubUpgradeId)
+            removeItem(caster, ITEM_MEDALLION_COURAGE)
+        else
+            simError(owner, "This unit already used the medallion of Courage")
+
+init
+    registerSpellEffectEvent(SPELL_SUPERSUB_COURAGE, () -> onCast())
+
+    // for i = 12 downto 0
+    //     SetPlayerAbilityAvailable(players[i], ABILITY_MEU_DUMMY, false)
```
theQuazz: Why is this commented out? We should either add a comment point out why or remove this code.

Jaccouille: Oh yeah I forgot to add a comment here, so this was done in the legacy code so I recoded it, however I didn't understand what it was for, so I commented it and saw no difference. Will add a comment.

Jaccouille:
Merged in #334

PR #333, opened 2019-03-23
Reworked spellbook

Jaccouille:
This branch is based off TrollUnitFactory, which is based off TrollUpgrade, so hum I think you should look at it after those two branch are merged, cause you can see all their commit & file change, I don't know if this was the correct way to proceed. Could use some advice.

No joke, even I can't tell which commit are from this branch ^^", except the commit I pushed on march 23.

Here's what I changed in this branch ->

1. Wurstified all spellBook.
2. Change some spells in sage and dementia spellbook, sage only has 1 inherited spellbook now.
3. Formatted autoSkill.wurst.
4. Small fixes on small details.


Jaccouille:
Merged in #334

PR #337, opened 2019-04-12
Modes quest

theQuazz:


wurst/systems/modes/GameMode.wurst
```diff
@@ -118,6 +118,24 @@ public class GameMode
 
                 modeboard.setTitle("Modes - 00:" + time)
 
+    static function createModeQuest()
```
jonathanmclaus: One quest per mode seems rather noisy. All this function should be plural I think to make its function more obvious.

theQuazz:
I do groups of 10

jonathanmclaus:
You're right, I missed the chunk. That seems better. Do you think it would make more sense to group modes, e.g. hard modes, easy mode, fun modes.

theQuazz:
Yeah that would be cool if someone wants to classify then

jonathanmclaus:
OK. I'll do that. Just going to add it to the constructor. Still a bit on
the fence about why we need a quest for modes but might as well do make it
nice.

theQuazz:
Alright I'll lay the groundwork in this PR and you can branch off of it and apply whichever categorization you'd like

theQuazz:
@jonathanmclaus updated cc @Jaccouille

Jaccouille:
@theQuazz let me know if you're gonna fix this branch since it's not based off master

theQuazz:
@Jaccouille updated

Jaccouille:
@theQuazz 
url

theQuazz:
@Jaccouille shoot! I'll take a look tonight

theQuazz:
@Jaccouille updated and tested (it works now)

Jaccouille:
url
I believe game mode quest are created following the game mode creation, first pro mode is standard, it is initialized before first casual mode, therefore, casual mode quest log is created after the pro one, do we care about that?

wurst/systems/modes/GameModeInit.wurst
```diff
@@ -0,0 +1,335 @@
...
+constant FAMINE                 = asList("famine"                , "fm"   , "Hard"   , "Animals spawn less")
+constant FAMINE_X2              = asList("famine-x2"             , "fm2"  , "Hard"   , "More intense famine")
+constant SLOW                   = asList("slow"                  , "sl"   , "Hard"   , "Spawn less items")
+constant LOWER_FOOD_LIMIT       = asList("lower-food-limit"      , "lfl"  , "Hard"   , "Lower spawn rate for animals on the map")
+constant LOWER_ITEM_LIMIT       = asList("lower-item-limit"      , "lil"  , "Hard"   , "Lower item spawn rate on the map")
+constant TEST_MODE              = asList("test-mode"             , "tm"   , "Testing", "No one can lose, enables test commands, type -spawn or -create to get explanation")
+constant OLD_ELEMENTALIST       = asList("old-elementalist"      , "oe"   , "Testing", "Replaces new elementalist with the old elementalist.")
+constant DEV                    = asList("dev"                   , "d"    , "Testing", "Development mode (-qp, -tm, -st)")
+constant TAMEABLE_FAWN          = asList("tameable-fawn"         , "tf"   , "Testing", "Tame fawns with acorns")
```
theQuazz: Hmm I'm not sure that I like these all broken out as lists rather than inline with the mode creation. WDYT @jonathanmclaus?

Jaccouille: Yeah I am just proposing that because I think inline is kinda annoying to read, I don't really like scrolling down

jonathanmclaus: I agree with Quazz. These are mode parameters and it doesn't make sense to separate arguments.

Jaccouille:
@theQuazz How's this?

wurst/systems/modes/GameModeInit.wurst
theQuazz: Let's remove this code here then I think we're good to go! Thanks for your work on this @Jaccouille

theQuazz:
I don't think that ordering is a huge problem however if we did want to enforce an order let me know and I will update this accordingly

Jaccouille:
@theQuazz Yeah I forgot to remove that, should be ok now

PR #341, opened 2019-05-11
Wrote UnitDefinition for animals

Jaccouille:
Not interesting, but it should be easier to update those later on


Jaccouille:
@theQuazz I copy/pasted definition from GUI, this might be a bit useless for pets since we will rework them,  I can see that marsun recoded some definition in his pet definition PR.

jonathanmclaus:
Is this good to accept now?

Jaccouille:
Yeah I believe it should be ok, just had a typo mistake

PR #345, opened 2019-06-18
Just cleaning up the import folder and fixing other small detail

Jaccouille:
Forgot to push commit from the 15 june+


Jaccouille:
Nevermind lol

PR #348, opened 2019-06-22
Fervor potion and small fixes

Jaccouille:
I reworked fervor potion because people said it's useless : 
-Gains 8 Damage
-Gains 6 Armor
-Gains 20% AS from unholy frenzy so it stacks with pump up
-Takes 9 Damage per seconds

Didn't give much thought about it, just wanted to try something in order to move on.

Changed WDH recipe to 3 stick 2 mana crystal, makes more sense to me.
Rest are small fixes.


theQuazz:
Sure if you want to give this a shot I'll approve but I think the main problem with fervor is that you can kite them

Jaccouille:
Oops, I pushed without checking if it was approved

PR #354, opened 2019-07-27
extracted constant files to /imports

Jaccouille:
I extracted the txt files for game & interfaces constants, those are imported into the map at compile time.

Would like to keep the build.w3x constant as it is, and extract the file each time we change the constant so we can keep track of the changes.

war3mapSkin.txt is kinda unreadable, but atleast we'll keep track of change.


jonathanmclaus:
I don't understand what you mean by keep the build.w3x constant at it is. Are you planning on keeping some of the constants in the map file instead of in text form?

Jaccouille:
> 
> 
> I don't understand what you mean by keep the build.w3x constant at it is. Are you planning on keeping some of the constants in the map file instead of in text form?

I wanted to mean that changing constant through the text form might be "un-practical" since we can't see all the constant field, and for the war3mapSkin.txt, the file itself doesn't seem readable to me since we got fields like "Inventoryfull=TRIGSTR_3461" and we can't really tell what "TRIGSTR_3461" refers to.

jonathanmclaus:
If you put the text file in imports, will it overwrite or conflict with the actual map file?

Frotty:
It will overwrite.
You can also get rid of using TRIGSTRs

jonathanmclaus:
Where are the TRIGSTR values defined?

Frotty:
war3map.wts, see https://github.com/inwc3/wc3libs/blob/master/src/main/java/net/moonlightflower/wc3libs/txt/WTS.java

jonathanmclaus:
But the `war3map.wts` isn't in this PR. So the definitions are already in the map binary and we can't see them?

And we can get rid of TRIGSTRs because Wurst will automatically write them?

Just wanted to make sure I understood before relying on it too much.

Frotty:
> But the war3map.wts isn't in this PR. So the definitions are already in the map binary and we can't see them?

The file is inside the map. You can open it with a text editor.

> And we can get rid of TRIGSTRs because Wurst will automatically write them?

Wurst does not currently replace TRIGSTRs, but I've written such a mechanism from my map optimizer which might just become a feature of wurst.
For now I just ment that you can get rid of them manually by inserting values instead of references.

jonathanmclaus:
So if we delete that file and then change all the references of TRIGSTR to actual strings, there won't be any problem? I wasn't sure if TRIGSTRs are required under-the-hood for interface constants.

Frotty:
You can replace all references and delete it, this is what vex's optimizer does.
For me it has also worked before, but many binary files contain references such as .w3i
I don't think removing the `.wts` file would be recommended for now.

jonathanmclaus:
@Frotty For removing TRIGSTR, do you write the out manually or parse the `.wts` file? Probably not worth building a parser if people don't use this normally, but I could if it's common enough.

jonathanmclaus:
Will revisit shortly.

PR #356, opened 2019-08-03
Drops raw meat now pick up additionnal raw meat if possible

Jaccouille:
At this moment, if you have for example, 3 raw meat picked up, to pick additional raw meat, you either have to click on the raw meat item or use Drop raw meat then Pick raw meat.

I believe having to use drop raw meat & pick raw meat to pick up additional raw meat is confusing.

I changed Drop raw meat so it will pick up additional raw meat if possible, if meat inventory is full or there's no meat around, it will drop the raw meat.

Since I am at it, I want to try another model for raw meat.


wurst/systems/meatsystem/MeatSystem.wurst
```diff
@@ -320,16 +325,16 @@ function onTargetOrder()
         and u.isInventoryFull()
         and (meatInventoryInstances.has(u) or isUnitMeatCarrier(u)))
 
-        MeatInventory instance = null
+        MeatInventory _instance = null
```
jonathanmclaus: Is `instance` a reserved word now? I'm not clear on why you add to add the underscore to the name.

Issue #349, opened 2019-07-05
Mammoth Horn can become unusable


jonathanmclaus:
The Mammoth Horn item opens a spellbook upon activation. There is a bug in WC3 that causes the spellbook for any item to fail to open if the item was moved within the inventory.

We should either block the item swap (with an error message telling them drop it) or issue commands to drop and pick up the item again. The latter is probably not worthwhile, as dummy items would be required in order to guarantee that the item ends up in the intended slot.

PR #358, opened 2019-08-19
Added workaround for spellbook failure on inventory move.

jonathanmclaus:


jonathanmclaus:
This should be viewed with whitespace off because I changed line endings.

jonathanmclaus:
Fixes #349

Jaccouille:
Why are the files fully highlighted when I look for diff on the github interface? [Files Changed](https://github.com/island-troll-tribes/island-troll-tribes/pull/358/files)

HallerJohn:
He might have copy pasted the file. Github treats that as, you deleted the whole page then added all of this new stuff

jonathanmclaus:
I touched the file in VS Code, which automatically changed all the line endings to the right ones. Like I said, view it without considering the whitespace in the diff.

jonathanmclaus:
Issue is that moving a spellbook in your inventory breaks it - you can't use the abilities and it can lock you inside the spellbook without you realizing. This reduces the problems associated with that by automatically replacing the spellbook in your inventory if you move it.

PR #362, opened 2019-08-23
Fixed interface error where inventory slots were blocked.

jonathanmclaus:


jonathanmclaus:
This is ready to accept now.

PR #368, opened 2019-09-05
Moved custom objects to Wurst.

jonathanmclaus:


Jaccouille:
Can you describe the goal/purpose, what did you remove?

jonathanmclaus:
The goal is not have any objects defined in the binary map file.

I copied the object editing output and move it to wurst, omitting any objects that were already defined in Wurst. I added custom object definitions for any base object that was modified.

I deleted all custom objects in the map file, but left the modified base objects alone, as there are custom objects that are implicitly reliant upon them.

PR #371, opened 2019-09-08
Removed deprecated Jurst assets.

jonathanmclaus:


jonathanmclaus:
I removed Camouflage coat because no one seems to make it and it would probably need to be changed rather than just migrated to Wurst.

PR #373, opened 2019-09-08
Update config.yml

Frotty:
Just checking if this has any effect on the output.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,6 @@ version: 2
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


Frotty:
why merge? I dont think this did anything 😅

jonathanmclaus:
@Frotty I don't think it uses this changes until it's merged to master, so I was just giving it a try. That's my understanding of how a `Jenkinsfile` works, but it might be different for Circle.

theQuazz:
It will run the version on the PR

PR #377, opened 2019-09-11
Removed Radar Scout.

jonathanmclaus:


jonathanmclaus:
Need to make sure this doesn't affect skill points for scout classes first.

PR #387, opened 2019-09-19
Added Forge, moved Workshop and Armory recipes accordingly.

HallerJohn:
Created Forge
Shifted recipes between forge/armory/workshop


wurst/systems/crafting/WorkshopQuickMake.wurst
```diff
@@ -50,6 +14,15 @@ init
     ..add(ITEM_TINDER)
     ..add(ITEM_TINDER)
 
+    new QuickMake(SPELL_QM_ENSNARE_TRAP)
+    ..add(ITEM_TINDER)
+    ..add(asList(ITEM_BONE, ITEM_STICK))
+    ..add(ITEM_STICK)
+
+    new QuickMake(SPELL_QM_MANA_CRYSTAL)
```
Jaccouille: You can craft mana crystal now?

Jaccouille:
Might want to rename "SlaughterHouse.mdx" to something more explicit.

wurst/_items/Kits/ForgeKit.wurst
```diff
@@ -0,0 +1,51 @@
...
+import ItemObjEditing
+import AbilityObjEditing
+import ObjectIds
+// import PublicLibrary
+import ID
+
+let ABILITY_ID    = compiletime(ABIL_ID_GEN.next())
+let ITEM_ID       = compiletime(ABIL_ID_GEN.next())
+let BASE_ITEM_ID  = 'rst1'
+let MODEL         = "Doodads\\LordaeronSummer\\Terrain\\Barricade\\Barricade0.mdx"
```
jonathanmclaus: Do not reference string constants like this directly. Either use the `Assets` package in the standard library or the `LocalAssets` package I recently added for our own imports. The `LivingClay` package demonstrats both.

jonathanmclaus: We don't have `ItemIds` in the standard library, so please add that to `LocalAssets`, even though it technically isn't local.

wurst/systems/crafting/WorkshopQuickMake.wurst
HallerJohn: yeah people said they wanted to be able to be able to craft some base components and i thought mana crystal was a good starting place

wurst/lib/LocalAssets.wurst
```diff
@@ -2,6 +2,7 @@ package LocalAssets
 
 public class LocalIcons
     static constant bTNGreenSentryWard = "ReplaceableTextures\\CommandButtons\\BTNGreenSentryWard.blp"
+    static constant btnForgeKit        = "ReplaceableTextures\\CommandButtons\\BTNNewForge.blp"
```
jonathanmclaus: Did you add this to `imports/`? I don't see it. I also don't think it makes much sense to call it "new".

wurst/systems/units/buildings/Armory.wurst
```diff
@@ -7,7 +7,7 @@ import UnitObjEditing
 import ObjectIds
 
 
-constant MODEL       = "Models\\SlaughterHouse.mdx"
+constant MODEL       = "Models\\Armory.mdx"
```
jonathanmclaus: Add a new class to `LocalAssets` called `LocalBuildings` and add this there, emulating https://github.com/wurstscript/WurstStdlib2/blob/master/wurst/_wurst/assets/Buildings.wurst#L4

wurst/lib/LocalAssets.wurst
jonathanmclaus: Realized I was only looking at the diff for the most recent commit when I couldn't find this. I see it now. You should still rename it though.

HallerJohn: yeah i just named it that because the old armory model was named forge and at the time i didnt want to erase anything. ill change that now though

PR #403, opened 2019-10-03
Added Forge Quickmake

HallerJohn:
Added Forge quickmake to trolls
Added Forge quickmake to master crafter
Removed armory quickmake in master crafter


wurst/systems/units/buildings/CraftMaster.wurst
```diff
@@ -22,7 +22,7 @@ public class CraftMaster extends Building
         getUnit().addAbility(SPELL_QM_STORAGE_HUT_KIT)
         getUnit().addAbility(SPELL_QM_HATCHERY_KIT)
         getUnit().addAbility(SPELL_QM_MUD_HUT_KIT)
-        getUnit().addAbility(SPELL_QM_ARMORY_KIT)
+        getUnit().addAbility(SPELL_QM_FORGE_KIT)
```
jonathanmclaus: Is the decision here to just not enforce that an armory can be created at the craft master? That seems fine because any troll can make the armory themself, just wanted to confirm.

HallerJohn: Yeah thas right

Issue #352, opened 2019-07-23
Observer Mode Board's broken


Jaccouille:
In observer mode, if you click on the board in multiplayer or lan, you get disconnected.

I investigated a bit and believe crash happens on this line : https://github.com/island-troll-tribes/island-troll-tribes/blob/master/wurst/systems/boards/ObserverBoard.wurst#L153

I don't know why

PR #408, opened 2019-10-12
Fixed multiboard bug caused by an observer leaving.

HallerJohn:
*Fix bug with scoreboard caused by obs leaving game.

### DIFF
--- a/wurst/systems/boards/TribeBoard.wurst
+++ b/wurst/systems/boards/TribeBoard.wurst
@@ -17,6 +17,7 @@ import Game
 import PlayerExtensions
 import StringExtensions
 import Tribe
+import EventHelper
 
 let PERIOD = 1.0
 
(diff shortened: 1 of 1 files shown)


Jaccouille:
Scoreboard? Isn't it a bug rather related to Tribeboard? Can you describe it? Is it related to #352? Did you test the fix?
You can test by opening 3 warcraft 3 client, create a LAN game, join it with all the client, start it, put 1 client in observer mode at the start of the game, then test the fix.

jonathanmclaus:
I think the terms scoreboard and multiboard are being conflated. This is referring to the board used to display the statistics for each member of your tribe, i.e. health, energy, and heat.

HallerJohn:
Yeah that was my mistake I mean the tribe board. The bug was that it would print an error message if an obs left the game. And yes me and mt tested it and it worked fine.

jonathanmclaus:
We didn't test if this fixes the issue in #352 but I doubt it's related.

PR #413, opened 2019-10-17
NE island pathing fix.

HallerJohn:
*Fixed pathing in top right of NE island.


HallerJohn:
image
This is the changed terrain

PR #416, opened 2019-10-21
Restored proper radius for merchant interaction.

jonathanmclaus:


jonathanmclaus:
No. It does affect gameplay but we've had no bug reports about the reduction in the radius the last month, so I don't think it's noticeable to the casual player.

PR #428, opened 2019-11-20
Alligator Man change, Ele nerf, minor bugfix.

HallerJohn:
$changelog: Elementalist's Brambles cooldown +25%
$changelog: Elementalist's Earth Guardian cooldown +25%
$changelog: Elementalist's Frost Blast cooldown +25%
$changelog: Alligator Man will now be invulnerable and regen health while not fighting on his island.
$changelog: The One will now also heal similarly to Alligator Man.
$changelog: Fixed bug where Elder Panther had Jungle Wolf abilities instead of its normal ones.
*Started changing function names to CamelCase.


wurst/systems/entities/bosses/AlligatorMan.wurst
```diff
@@ -2,7 +2,30 @@ package AlligatorMan
 import ID
 import Hostile
 import LinkedList
+import BossUtilities
+import RegisterEvents
+import HashMap
 
+//Regions for the alligator islands
+public let NW_ALLIGATOR_REGION = CreateRegion()
```
jonathanmclaus: You don't reference these further than creating the overall list, so I would just define them inline there to simply it.

    let regions = asList(
        CreateRegion..addRect(gg_rct_NW_ALLIGATOR),
        ...
    )

wurst/systems/entities/bosses/AlligatorMan.wurst
```diff
@@ -2,7 +2,30 @@ package AlligatorMan
...
+//List of alligator regions
+let REGIONS = asList(
+    NW_ALLIGATOR_REGION,
+    NE_ALLIGATOR_REGION,
+    SW_ALLIGATOR_REGION,
+    SE_ALLIGATOR_REGION
+)
+
+//Mapping of alligator regions to it's corresponding alligator
+var regionsToAlligators = new IterableMap<region, unit>()
```
jonathanmclaus: Does this not work as `let`? You need `var` to reassign the variable, but I think that modification via methods works fine with `let` because the reference itself is constant.

wurst/systems/entities/bosses/AlligatorMan.wurst
```diff
@@ -26,6 +49,42 @@ public class AlligatorMan extends Hostile
     override function getNumCorpses() returns int
         return 10
 
-
 init
     AlligatorMan.initialize()
+
+    //Cycle through alligator regions
+    for curr_region in REGIONS
```
jonathanmclaus: The manual states that variable names `lookLikeThis`.

wurst/systems/entities/bosses/AlligatorMan.wurst
jonathanmclaus: I'm on the fence about whether this is a constant with regards to using uppercase for it. You may want to modify it later.  I typically only use uppercase for simple types that can't be modified at all, like integers and strings.

wurst/systems/entities/bosses/AlligatorMan.wurst
```diff
@@ -26,6 +49,42 @@ public class AlligatorMan extends Hostile
...
+        //Create trigger for detecting alligator
+        let x =CreateTrigger()
+        x.registerEnterRegion(curr_region, null)
+        x.addAction()->
+            let trigger_region = GetTriggeringRegion()
+            let entering_unit = GetEnteringUnit()
+            //Verify entering unit is alligator
+            if entering_unit.getTypeId() == UNIT_ALIGATOR_MAN
+                //Check to make sure we don't already have that alligator
+                if not regionsToAlligators.has(trigger_region)
```
jonathanmclaus: You don't need this check because you destroy the trigger before this ever evaluates to false.

wurst/systems/entities/bosses/AlligatorMan.wurst
```diff
@@ -2,7 +2,30 @@ package AlligatorMan
 import ID
 import Hostile
 import LinkedList
+import BossUtilities
```
jonathanmclaus: Please separate out imports. It makes it easier to visualize what we have control over.

wurst/systems/entities/bosses/BossUtilities.wurst
```diff
@@ -0,0 +1,46 @@
...
+    for troll in getTrolls()
+        // Disable invulnerability if both conditions are met.
+        if troll.isAlive() and troll.isInRegion(mini_Boss_Region)
+            return false
+
+    // Enable invulnerability if no matches were found.
+    return true
+
+public function toggleInvulnerability(region mini_Boss_Region, unit mini_Boss)
+            // Exit if The Boss is not currently alive.
```
jonathanmclaus: Fix this indent.

wurst/systems/entities/bosses/TheOne.wurst
```diff
@@ -82,40 +81,19 @@ public function spawnTheOne()
...
-        ..addAction(function toggleInvulnerability)
+        ..addAction()->
+        if theOne != null
+            toggleInvulnerability(THE_ONE_REGION, theOne.getUnit())
     registerPlayerUnitEvent(
-        EVENT_PLAYER_UNIT_DEATH,
-        function toggleInvulnerability
-    )
+        EVENT_PLAYER_UNIT_DEATH)->
+        if theOne != null
```
jonathanmclaus: I would leave the check for null in `toggleInvulnerability`. It's cheap and simplifies the logic for conditional bosses.

wurst/_units/Animals/PantherDefinition.wurst
```diff
@@ -12,6 +12,7 @@ constant PANTHER_ICON = "ReplaceableTextures\\CommandButtons\\BTNPanther.blp"
 class ElderPantherDefinition extends JungleWolfDefinition
     construct(int unitId)
         super(unitId)
+        this.setNormalAbilities(PANTHER_ABIL)
```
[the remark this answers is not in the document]
HallerJohn: the current class hierarchy is Jungle Wolf -> Elder Panther -> Panther. So panther would inherit those abilities from elder.

wurst/systems/entities/bosses/TheOne.wurst
HallerJohn: right now toggleInvulnerability is given a unit and region, doesnt it make sense to keep that check outside the function since not all bosses we will be checking have an object like that?

jonathanmclaus: It doesn't complicate the logic and is simpler to write once. Plus it works as a safeguard against future mistakes where you might forget to check for null otherwise.

HallerJohn: mmk, what would i use to accept any class? ie. TheOne or DiscoDuck

jonathanmclaus: It would still take a `unit`. I just mean that you should add the null check there. `unit` can also be null.

jonathanmclaus: Ah, I see what the problem is. You would still need the null check on `theOne` before calling `getUnit`. I would add a second function that takes `UnitEntity` and just forwards the call to the function you have now if the entity isn't null.

wurst/systems/entities/bosses/BossUtilities.wurst
```diff
@@ -1,46 +1,54 @@
...
-import GroupUtils
 import ClosureTimers
+
+// Local imports:
 import HealingSystem
+import GroupUtils
+import UnitEntity
 
-public function checkInvulnerability(unit mini_Boss, region mini_Boss_Region) returns bool
+public function checkInvulnerability(region miniBossRegion, unit miniBoss) returns bool
```
jonathanmclaus: I don't think that there's really a distinction when it comes to miniboss or a reason why this system would be limited to that. I would omit the `mini` from all names, although implying that this is only for bosses should be fine.

wurst/systems/entities/bosses/AlligatorMan.wurst
```diff
@@ -26,6 +44,40 @@ public class AlligatorMan extends Hostile
...
-
 init
     AlligatorMan.initialize()
+
+    // Cycle through alligator regions
+    for currentRegion in regions
+        // Create trigger for detecting alligator
+        let x =CreateTrigger()
+        x.registerEnterRegion(currentRegion, null)
+        x.addAction()->
```
jonathanmclaus: Put spaces around binary operators, including `->`.

wurst/systems/entities/bosses/AlligatorMan.wurst
```diff
@@ -26,6 +44,40 @@ public class AlligatorMan extends Hostile
...
 
-
 init
     AlligatorMan.initialize()
+
+    // Cycle through alligator regions
+    for currentRegion in regions
+        // Create trigger for detecting alligator
+        let x =CreateTrigger()
+        x.registerEnterRegion(currentRegion, null)
```
jonathanmclaus: Don't assign this trigger, just use `..` notation.

    CreateTrigger()
        ..registerEnterRegion(...)
        ..addAction() ->
            // some code here

wurst/systems/entities/bosses/TheOne.wurst
```diff
@@ -82,40 +81,19 @@ public function spawnTheOne()
...
-    else
-        theOne.getUnit().removeAbility(AbilityIds.invulnerable)
-
 init
     TheOne.initialize()
     CreateTrigger()
         ..registerEnterRegion(THE_ONE_REGION, null)
-        ..addAction(function toggleInvulnerability)
+        ..addAction()->
+        if theOne != null
```
jonathanmclaus: This should be indented. Not sure if it works as is but it's easier to follow if indented.

wurst/systems/entities/bosses/BossUtilities.wurst
```diff
@@ -0,0 +1,54 @@
...
+    // Heal the boss 4 hp every 1 second
+    doPeriodically(regenPeriod) (CallbackPeriodic cb) ->
+        // Use healing system
+        new HealingInstance(boss, hpRegen, HealingType.OTHER)
+        // If boss is not invulnerable or is max health
+        if not boss.isInvulnerable() or boss.getMissingHP() == 0
+            // Stop healing
+            destroy cb
+
+public function checkEntity(UnitEntity bossEntity, region bossRegion, unit boss)
```
jonathanmclaus: You don't need to pass `boss` directly. Just use `bossEntity.getUnit()`. Once you don't need that identifier, you can use `boss` instead of `bossEntity` as the entity is implied by the type.

Also, you should be able to name this function `toggleInvulnerability` because Wurst allows function overloading.

wurst/systems/entities/bosses/BossUtilities.wurst
```diff
@@ -0,0 +1,54 @@
...
+        // Disable invulnerability if both conditions are met.
+        if troll.isAlive() and troll.isInRegion(bossRegion)
+            return false
+
+    // Enable invulnerability if no matches were found.
+    return true
+
+public function toggleInvulnerability(region bossRegion, unit boss)
+    // Exit if The Boss is not currently alive.
+    if not boss.isAlive()
```
jonathanmclaus: Check that boss is not null as well, otherwise this will throw an error.

PR #439, opened 2020-01-18
Fixed various bugs.

Jaccouille:
$changelog: Fixed bug where using a Cure All potion would not be visible or remove some conditions.
$changelog: Lowered the hit points for the Elk in the demo so that it dies quickly.
$changelog: Removed inconsistency in models for different types of spears.

I rewrote cure all item definition & dummy ability in wurst, might want to remove the old definition from https://github.com/island-troll-tribes/island-troll-tribes/blob/88afb5a033b2ba1e569325109c2d2a3c50ccf5a6/wurst/objects/items/Raw.wurst#L2236


wurst/objects/items/MixingPot/Cure.wurst
```diff
@@ -0,0 +1,38 @@
...
+    new ChannelAbilityPreset(ABIL, 1, true)
+    ..setName("Cure Potion")
+    ..setCheckDependencies(false)
+    ..setHeroAbility(false)
+    ..setItemAbility(true)
+    ..presetTargetTypes(Targettype.UNIT)
+    ..setManaCost(1, 0)
+    ..setCooldown(1, 0)
+    ..setCastRange(1, CAST_RANGE)
+    ..setEditorSuffix("(Wurst)")
```
jonathanmclaus: Don't do this. Wurst does it automatically and doing it manually is just clutter.

wurst/objects/items/MixingPot/Cure.wurst
```diff
@@ -0,0 +1,38 @@
...
+
+constant ABIL  = compiletime(ABIL_ID_GEN.next())
+
+//constant ABIL  = "A05K"
+constant ICON  = ICON_PATH.format("LesserClarityPotion")
+constant MODEL = "Models\\PotionGreenSmall.mdl"
+
+constant CAST_RANGE = 500.
+constant TARGET_ALLOWED = "friend,organic,self"
+constant TT = "This potion is said to be able to cure anything. Made from Athelas Seed."
```
jonathanmclaus: Don't mention ingredients. That should follow from the recipe and suggests an inverse relationship.

wurst/objects/items/MixingPot/ManaPotion.wurst
```diff
@@ -33,9 +33,9 @@ constant TARGET_ALLOWED = "friend,ground,hero,invulnerable,organic,self,vulnerab
...
 
 @compiletime function createManaPotion()
-    new PotionDefinition(ITEM_MANA_POTION_I, ICON, MODEL, "Mana Potion", TT, 5, asList(255, 255, 255), commaList(ABIL), 0.7, "A002", 120)
+    new PotionDefinition(ITEM_MANA_POTION_I, ICON, MODEL, "Mana Potion", "W", TT, 5, commaList(ABIL), 0.7, "A002", 120)
 
 init
     EventListener.onTargetCast(ABIL) (unit caster, unit target) ->
         target.addMana(MANA_RESTORED)
-        flashEffect(Abilities.aImaTarget, target, "origin")
+        flashEffect(Abilities.aImaTarget, target, "origin")
```
jonathanmclaus: Can you fix your editor not to strip EOF linebreaks?

wurst/systems/items/spellbooks.wurst
```diff
@@ -11,6 +11,7 @@ import InventoryEvent
 import ItemExtensions
 import UnitExtensions
 
+// This is a fix for the Horn of Mammoth ability bug, sometime when you click on the horn, there is no ability available
```
jonathanmclaus: Don't refer to Mammoth of Horn directly. Can you just describe the general bug for spellbook items?

wurst/__legacy__/lib/ID.jurst
```diff
@@ -882,7 +882,8 @@ library ID requires ObjectIdGenerator, AdvanceIDs
 		public constant integer ITEM_ACID_BOMB               = 'I04K'
 		public constant integer ITEM_ANTI_MAGIC_POTION       = 'I02M'
 		public constant integer ITEM_ANABOLIC_POTION         = 'I02H'
-		public constant integer ITEM_CURE_ALL                = 'I03G'
+		public constant integer ITEM_CURE_ALL                = compiletime(ITEM_ID_GEN.next())
+		// public constant integer ITEM_CURE_ALL                = 'I03G'
```
jonathanmclaus: Just delete commented code. You can use the git history if you really need to see the old version.

wurst/objects/items/MixingPot/Cure.wurst
```diff
@@ -0,0 +1,38 @@
...
+
+import PotionsDefinition
+import ID
+import ChannelAbilityPreset
+import ClosureEvents
+import InstantDummyCaster
+import Orders
+import ObjEditingUtils
+
+constant ABIL  = compiletime(ABIL_ID_GEN.next())
```
jonathanmclaus: Abilities should be globally registered in a single file to prevent imports just for the IDs.

Jaccouille: Even for item dummy abilities? I don't see a scenario case where you would need to import it since it will(should) be used only for this item.

wurst/objects/items/MixingPot/Cure.wurst
Jaccouille: Wurst does it automatically? I added this suffix to make sure the spell I checked in the GUI editor was made in wurst, I believed it would be clearer, well at least for me

jonathanmclaus: Wurst sets its a custom field that isn't visible from the object editor, but is included in the object editing output. Are you opening the built map in the World Editor to debug?

Jaccouille: Yes, in some case I used the World Editor to debug, I took this habit when I started recoding the troll spells, I didn't remove the GUI spell at first, so I added the suffix to make a diff between old and new spell in the GUI.

wurst/objects/items/MixingPot/Cure.wurst
jonathanmclaus: Well in this case it was the dummy ability that the item uses that was the problem, so testing those directly doesn't seem like a bad ID. Having a single directory of IDs could facilitate having commands to test abilities directly.

wurst/systems/items/spellbooks.wurst
Jaccouille: "This is a fix for opening a spellbook on item click, sometime the spellbook would open but didn't show its abilities, in this case, the unit needed to drop the item and pick it up again"

Is this ok?

wurst/systems/crafting/QuickMakeToolTips.wurst
```diff
@@ -9,44 +9,44 @@ import StringExtensions
...
 // Herbs
-constant let SPECIAL_HERB       = "Special Herb".color(SPECIAL_COLOR)
-constant let BLUE_HERB          = "Blue Herb".color(COLOR_BLUE)
-constant let ORANGE_HERB        = "Orange Herb".color(COLOR_ORANGE)
-constant let PURPLE_HERB        = "Purple Herb".color(COLOR_PURPLE)
-constant let YELLOW_HERB        = "Yellow Herb".color(COLOR_YELLOW)
-constant let RIVER_ROOT         = "River Root".color(COLOR_TEAL)
-constant let RIVER_STEM         = "River Stem".color(COLOR_GRAY)
-constant let ATHELAS_SEED       = "Athelas Seed".color(COLOR_GREEN)
+constant SPECIAL_HERB       = "Special Herb".color(SPECIAL_COLOR)
```
jonathanmclaus: We use `let` a lot more often than `constant` and they mean the same thing, so can you keep `let` instead of `constant`?

Jaccouille: Kinda prefer `constant`over `let` because it seems more explicit to me but okay.

jonathanmclaus: Yeah, I'd prefer type modifiers over separate keywords but `let` seems to be preferred throughout Wurst.

Issue #434, opened 2019-12-13
Popping ward does not affect enemy


jonathanmclaus:
An enemy walking over a ward will no longer play the animation for the enemy. Their unit will be damaged and flash on the minimap, but will not directly see the ward explode, making it harder to notice.

jonathanmclaus:
Closed by #440

PR #440, opened 2020-01-25
Fixed Living Clay Invisibility

Jaccouille:
Current living clay does not become visible when dying, if you walk over it, it disappear and deals damage, but the player cannot see it, if you reveal them and walk over them, you can see them dying.

Old living clay has the Abilities.ghostVisible 'Aeth', current living clay has Abilities.ghost 'Agho', I quickly tested and it seemed to be the fix, haven't found documentation on the difference between those abilities yet though.

Can you tested on your side and confirm?

Question : Any idea why do I have the commits displayed below when there is no change in them? I think it happened after I updated the project with wurst setup.

Note : Look like the ghost abilities isn't required to make the unit invisible, goblin landmine is invisible by default, guess it's hardcoded. However Sentry Ward has the ghostVisible ability, so we'll still include it just in case, ghost makes so that you ignore the ward when pathing.


jonathanmclaus:
Fixes #434

PR #442, opened 2020-01-27
Migrated map preview file from binary to Wurst

jonathanmclaus:
$changelog: Changed preview to Zandalari crest.


jonathanmclaus:
Undid this via rewriting the Git history because the message had a typo.

Issue #438, opened 2019-12-31
Order IDs from factory for Channel conflict with basic abilities


jonathanmclaus:
In v3.5c, an Elementalist cannot cast Cripple because the order ID for transforming into Dementia Master uses the same order ID. This may disappear in random builds because the factory generating order IDs for channel abilities like the transformation abilities is non-deterministic.

jonathanmclaus:
Closed by #445

PR #445, opened 2020-01-29
Corrected conflicting order IDs.

jonathanmclaus:
$changelog: Fixed bug where issuing an order would sometimes use the wrong ability.

Also adds more features to the toolkit.


jonathanmclaus:
Closes #438

PR #446, opened 2020-01-29
Quick update to make the game playable on reforged client

Mephistophelesss:
$changelog: Removed Stone and Mana Crystal Stacking.
$changelog: Changed Mana Crystal Model (Temporarily).


Mephistophelesss:
apparently fire and other construction buildings might also stack

PR #451, opened 2020-01-30
Fixed custom models.

Jaccouille:
$changelog : Fixed display issues for Horn of the Mammoth and Essence of Darkness.

Horn of mammoth & orb of darkness used a texture that has been removed with latest patch, I changed the texture used by the models to fix it.


jonathanmclaus:
Is the texture built into the model? I thought it was separate but it looks you're just updating the models in place.

Jaccouille:
> 
> 
> Is the texture built into the model? I thought it was separate but it looks you're just updating the models in place.

The model file (mdx) got a path to a texture file (blp before, dds now), I just changed this path to an existing texture file using a third party software

PR #460, opened 2020-02-03
Fixed loading screen for Reforged client.

Jaccouille:
$changelog: Restored custom loading screen.
$changelog: Added Discord link in map description.

We used to overwrite the files in ui/glues/loading/load-generic/ and select the generic loading screen through world editor.

Problem : With reforged update, the game client do not use the loading screen contained in ui/glues/loadingscreen/ at all, the reforged client uses the files located at webui/loadingscreen/, those are jpeg files.

I tried to overwrite those files and failed, so I used an imported model, this model holding the loading screen as a texture.

I edited base.w3x, manually importing the LoadingScreen.mdx model which is used to display the loading screen. I then selected the model in the "Use imported loading screen" input.

To update the loading screen, you have to change the LoadingScreen.tga file in imports/
I used a third party software called [BLP Lab](https://www.hiveworkshop.com/threads/blp-lab-v0-5-0.137599/) to create this file.

Start BLP Lab, open a png file or jpeg then click on save as, choose the tga format and a filename (LoadingScreen.tga in our case) then click on "OK"

The LoadingScreen.mdx is a model template for loading screen, I got it from another map make, it doesn't need any editing.

To avoid having the map name & description displayed on the loading screen, I had to put a space in the title, subtitle and text field from the scenario loading screen data.


wurst.build
```diff
@@ -10,12 +10,13 @@ buildMapData:
   scenarioData:
     description: Island Troll Tribes is a PvP-centric survival map in which multiple
       tribes of trolls compete to gather resources, craft equipment, and eradicate
-      each other. Join us at https://discord.gg/hYzTRYf.
+      each other. Join our Discord server on https://discord.gg/Rh9JdKs for mustering lobbies, reporting bugs, and making
```
jonathanmclaus: I think it should be `at https...`. Also, did Wurst not automatically resize this when running? I left mine as a really long line and it formatted the description into 3 lines instead.

Jaccouille: I noticed this too, but it looks like it doesn't resize the lines on my side.

jonathanmclaus:
Can you describe of process of creating the `.tga`  in case someone else wants to update in the future? Did you modify the `.mdx` from the template at all or is just modifying the `.tga` needed?

Jaccouille:
> 
> 
> Can you describe of process of creating the `.tga` in case someone else wants to update in the future? Did you modify the `.mdx` from the template at all or is just modifying the `.tga` needed?

Updated initial description, just modifying the `.tga` is required.

PR #466, opened 2020-02-05
Reworked Medallion of Courage implemenation.

Jaccouille:
$changelog: Fixed Medallion Of Courage icon bug

Removed Tinker engineer upgrade due to recent bug on reforged, more over I didn't find the tutorial I used to follow on HiveWorkshop, nor remember why we used it, was it to add an hero ability which didn't exist before?

I didn't manage to make the ability require 2 upgrade, so I ended up using only 1

Also remapped the sub hotkey to fit QWE instead of WER, feels more consistent with button position to me but that's just a detail.


jonathanmclaus:
Yeah, I don't think Engineering is even meaningful. It was used to hide information about early supersub from players, but I don't think that's good for gameplay. Hopefully mentioning the Medallion of Courage won't confuse players who have never seen one.

Apparently Omnigatherer is easy...

wurst/objects/items/MedallionOfCourage.wurst
```diff
@@ -80,17 +24,13 @@ function onCast(unit caster, item target)
...
-            string medalHint = "{0} {1}".format(caster.getProperName().color(GOLD_COLOR), "acquired a super class choice! Check your hero skill and reach level 9 to use it.".color(GENERAL_COLOR))
-            printTimedToPlayer(medalHint, 5, owner)
-            SetPlayerAbilityAvailable(owner, _superSubUpgradeId, false)
-            caster.addAbility(_superSubUpgradeId)
-            target.remove()
-        else
-            simError(owner, "This unit already used the medallion of Courage")
+    if GetPlayerTechResearched(caster.getOwner(), UPGD_MEDALLION_COURAGE, true) == false
+        SetPlayerTechResearched(caster.getOwner(), UPGD_MEDALLION_COURAGE, 1)
+        let medalHint = "{0} {1}".format(caster.getProperName().color(GOLD_COLOR), "acquired a super class choice! Check your hero skill and reach level 9 to use it.".color(GENERAL_COLOR))
```
jonathanmclaus: This is better written as `"{0} acquired a supersub class choice...".format(caster.getProperName...`

Basically, just include the hardcoded part of the string as part of the original format string, not as an argument.

wurst/objects/items/MedallionOfCourage.wurst
```diff
@@ -80,17 +24,13 @@ function onCast(unit caster, item target)
...
+        printTimedToPlayer(medalHint, 5, owner)
+        target.remove()
+    else
+        simError(owner, "This unit already used the medallion of Courage")
```
jonathanmclaus: You have inconsistent capitalization here.

wurst/lib/Transformation.wurst
```diff
@@ -113,5 +115,10 @@ public function transformUnit(unit origin, int targetID)
         // Apply any callbacks registered as succeeding the transformation.
         runCallbacks(after, origin, originID)
 
+        // Once the unit transform into a subclass, medallion of courage is researched
```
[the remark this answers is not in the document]
Jaccouille: My bad, moved it to TrollUpgrade.

wurst/objects/items/MedallionOfCourage.wurst
```diff
@@ -80,17 +24,13 @@ function onCast(unit caster, item target)
...
-            string medalHint = "{0} {1}".format(caster.getProperName().color(GOLD_COLOR), "acquired a super class choice! Check your hero skill and reach level 9 to use it.".color(GENERAL_COLOR))
-            printTimedToPlayer(medalHint, 5, owner)
-            SetPlayerAbilityAvailable(owner, _superSubUpgradeId, false)
-            caster.addAbility(_superSubUpgradeId)
-            target.remove()
-        else
-            simError(owner, "This unit already used the medallion of Courage")
+    if GetPlayerTechResearched(caster.getOwner(), UPGD_MEDALLION_COURAGE, true) == false
+        SetPlayerTechResearched(caster.getOwner(), UPGD_MEDALLION_COURAGE, 1)
+        let medalHint = ("{0} acquired a super class choice! Check your hero skill and reach level 9 to use it.".color(GENERAL_COLOR)).format(caster.getProperName().color(GOLD_COLOR))
```
jonathanmclaus: I this might mess up the color though, because `{0}` will close the color tag for both the gold and general color.

wurst/systems/commands/SpawningTestHelpers.wurst
```diff
@@ -75,6 +75,18 @@ let SPAWN_HELP_MESSAGE = compiletime(
     ).joinBy("\n")
 )
 
+let RESEARCH_HELP_MESSAGE = compiletime(
```
jonathanmclaus: All of this should go into the toolkit (current `lib/commands`), but I can move it there if you'd prefer. I also think you should check the type of the ID directly instead of what it starts with, but I can demonstrate that when I move it.

The reason I wanted you to make it generic is because I plan on moving that lib into its own dependency to separate out ITT-specific code versus generic libraries.

wurst/objects/abilities/TrollUpgrade.wurst
```diff
@@ -95,6 +96,11 @@ function upgradeUnit(unit origin, int targetID)
     let agi = origin.getAgi(false)
     let inT = origin.getInt(false)
 
+    // Once the unit transform into a subclass, medallion of courage is researched
+    // this way the super sub spell gets unlocked
+    if UNSUB_TROLL_UNIT_LIST.has(origin.getTypeId()) and GetPlayerTechResearched(origin.getOwner(), UPGD_MEDALLION_COURAGE, true) == false
```
jonathanmclaus: I don't really see the point of this check. Whenever a troll sub-classes, it should always have the medallion upgrade. Having a check would make sense if there were more than three tiers of classes, but I doubt we need to worry about that.

Also, checking for the research is pointless. Just research it anyway (unless that causes a bug or something). So at the very least you only need the check on the `UNSUB_TROLL_UNIT_LIST`.

wurst/systems/commands/SpawningTestHelpers.wurst
Jaccouille: I kind of see what you're planning, but I'll let you demonstrate if you don't mind.

PR #467, opened 2020-02-06
Fixed item stacking.

Jaccouille:
$changelog: Reverted behavior for stacking and splitting of items.
$changelog: Fixed bug where Omnicure Potion would not have an effect.

So I understood your point about the way I rewrote Item definition with the long ass constructor, but since this code (spear one and omnicure) was already done on this branch https://github.com/Jaccouille/island-troll-tribes/tree/WurstifyItemDefinition/wurst/_items and is working, I decided to re-use in order to gain time, my priority being resolving most bugs asap.


wurst/assets/LocalAssets.wurst
```diff
@@ -31,10 +31,14 @@ public class LocalUnits
     // A nonexistent model used to hide units.
     static constant dummy = "Dummy.mdl"
 
-public class LocalBuildings
-    static constant forge          = "Models\\Forge.mdx"
-    static constant armory         = "Models\\Armory.mdx"
-    static constant mixingPot      = "Models\\MixingPot.mdl"
+public class LocalModelBuildings
```
jonathanmclaus: I would avoid mentioning `Model` to match the standard library assets. Having 2 classes is good though.

wurst/objects/items/MixingPot/OmnicurePotion.wurst
```diff
@@ -0,0 +1,19 @@
...
+
+constant ICON = ICON_PATH.format("MinorRejuvPotion")
+constant MODEL = LocalModelItems.potionPurpleSmall
+constant TT = "This potion is said to be able to cure all allies."
+
+@compiletime function createOmnicurePotion()
+    new PotionDefinition(ITEM_OMNICURE, ICON, MODEL, "Omnicure Potion", "S", TT, 22, commaList(ABILITY_OMNICURE_ITEM), 0.70, commaList(ABILITY_OMNICURE_ITEM), 120)
+
+init
+    registerSpellEffectEvent(ABILITY_OMNICURE_ITEM, function castOmnicure)
```
jonathanmclaus: This really should need to depend on the other `Omnicure` package directly. Use a dummy caster.

wurst/objects/abilities/Omnicure.wurst
```diff
@@ -40,19 +40,24 @@ class Omnicure extends ChannelAbilityPreset
...
+        this.setButtonPositionNormalX(0)
+        this.setButtonPositionNormalY(0)
+
 @compiletime function createOmnicure()
-    new Omnicure(ABILITY_OMNICURE, "E", new Pair(1, 0))
+    new Omnicure(ABILITY_OMNICURE, "E", new Pair(1, 0), MANACOST)
+    new Omnicure(ABILITY_OMNICURE_ITEM, "E", new Pair(1, 0), 0)
+
+public function castOmnicure()
+    let heroes = GetUnitsInRectMatching(bj_mapInitialPlayableArea, Condition(() -> begin
```
jonathanmclaus: Can you not do this in a less archaic way? Using filters, condition, and code in Wurst seems silly. `ClosuresForGroups` should replace this.

wurst/objects/items/MixingPot/OmnicurePotion.wurst
jonathanmclaus: If the casting is currently a problem, then I think the solution should be the same as Cure All, where we delay the recursive `InstantDummyCaster` usage.

jonathanmclaus:
Could you explain a bit about how this fixes stacking? I know that stack and split library was working for items that were already stackable, but it's not clear to me what the limitating are exactly with the new field and what we should be doing long term.

Also, meat had a problem with splitting and was already part of the stack and split library, so how does this fix splitting?

wurst/assets/LocalAssets.wurst
Jaccouille: Thought it would be more explicit this way

Jaccouille:
> 
> 
> Could you explain a bit about how this fixes stacking? I know that stack and split library was working for items that were already stackable, but it's not clear to me what the limitating are exactly with the new field and what we should be doing long term.
> 
> Also, meat had a problem with splitting and was already part of the stack and split library, so how does this fix splitting?

So I didn't want to dive into StackNSplit code, I noticed that most of our item were based off health stone, with reforged update, a new field called "Max Stack" has been added, health stone max stack is set to 5 by default, so our item max stack were set to 5 after reforged patch, hence why we can stack building kit now.

Splitting still work, but only when the stack charge is > to Max stack attribute
e.g, cooked meat can stack up to 10 with stackNsplit, but its Max Stack field is set to 5, we end up being able to split meat only when the stack charge is > 5, if you had 10 meat, you could split down to 5 meat.

Shouldn't take too much time to update stackNsplit but I really wanted a quickfix, so I setted Max Stack to 1 for all items.

jonathanmclaus:
OK. Wasn't aware of that. I thought I had tried it with a 10 stack and it still hadn't worked.

Don't worry about stack and split. I don't want to use it long term because it doesn't handled queued orders as best as possible. I'd rather create a new system that uses runes as fake items so you can queue as many commands possible.

wurst/objects/abilities/Omnicure.wurst
Jaccouille: Updated

wurst/objects/items/MixingPot/OmnicurePotion.wurst
Jaccouille: > 
> 
> This really should need to depend on the other `Omnicure` package directly.

I don't think I understand this sentence.

wurst/assets/LocalAssets.wurst
jonathanmclaus: I think it's clear in context because it's obvious when you're using a model, so just sticking with the standard convention is better.

wurst/objects/items/MixingPot/OmnicurePotion.wurst
jonathanmclaus: Oops, I meant should not. Just use a dummy caster so that the potion casts the real omnicure.

wurst/objects/abilities/Omnicure.wurst
jonathanmclaus: Can you use the `nullTimer` for the dummy cast, similar to how cure all was fixed.

wurst/objects/abilities/Omnicure.wurst
```diff
@@ -40,19 +46,11 @@ class Omnicure extends ChannelAbilityPreset
...
-
-        let owner = GetSpellAbilityUnit().getOwner()
-        for hero in heroes
-            new DummyCaster()
-                ..owner(owner)
-                ..origin(hero.getPos() + vec2(100, 0))
-                ..castTarget(ABILITY_CURE_ALL, 1, Orders.autodispel, hero)
+    EventListener.onCast(ABILITY_OMNICURE) (unit caster) ->
+        let owner = caster.getOwner()
+        forUnitsInRect(bj_mapInitialPlayableArea) (unit u) ->
```
jonathanmclaus: This needs a `nullTimer` doesn't it? Does the omnicure potion work with this current implementation? I would think it runs into the same problem that Cure All had where the banish wasn't being cast.

wurst/objects/abilities/Omnicure.wurst
```diff
@@ -40,19 +46,11 @@ class Omnicure extends ChannelAbilityPreset
...
+            if u.isType(UNIT_TYPE_HERO) and u.getOwner().isAllyOf(owner)
+                InstantDummyCaster.castTarget(owner, ABILITY_CURE_ALL, 1, Orders.autodispel, u)
```
jonathanmclaus: This is the wrong order ID. Not sure if you can use the compiletime definition.

jonathanmclaus:
OK, I think I was going about this the wrong way because you're latest commit still has the dependency between files and there's not a good way to avoid that with the current solution.

I would define both abilities in the same file and have them work exactly the same other than the mana cost. In particular, they'll both have the same event listener.

Then I would just have the potion use the ability ID of the 0 cost one. It won't depend on that file at all with that.

wurst/objects/abilities/Omnicure.wurst
Jaccouille: Wrong order ID? You mean it should be abolishmagic? Funny I didn't edit this dummyCast, but it's still working, gonna change to abolish just to make sure.

wurst/objects/abilities/Omnicure.wurst
Jaccouille: Yeah, I tested and it worked.

wurst/objects/abilities/Omnicure.wurst
Jaccouille: Well cure all use autodispell and there's no order for abolish magic.

alexandreborgo: Yeah you're right

jonathanmclaus: Confused myself. I was thinking this was where the potion was dummy casting the priest omnicure, not the priest omnicure casting cure all.

jonathanmclaus:
What happens if you use 0 instead of 1?

Jaccouille:
> 
> 
> What happens if you use 0 instead of 1?

I didn't notice any difference, apparently not being able to un-stack item is a known.
Is there any problem left with this PR?

Issue #468, opened 2020-02-08
Hydra doesn't spawn when summoned.


Jaccouille:

Jaccouille:
Has been fixed in #472

PR #472, opened 2020-02-08
Fixed hydra summoning.

Jaccouille:
$changelog: Fixed Hydra summoning

Issue #468
I believe disco duck region wasn't moved in #420, moved it into the red circle.
I thought meat had to be placed in the red circle to summon the hydra, but it doesn't work anymore, maybe altar has been placed too far during terrain rework.

IIRC the code, meat have to be placed around the altar (400 range), however, I never understood that, always thought it was in the red circle, so I believe it's clearer is we change the placement to the center of the red circle.

Green circle : Disco Duck region was set there
Red circle : Moved Disco Duck region to there


jonathanmclaus:
Can you include a screenshot?

Jaccouille:
Updated altar angle:

PR #479, opened 2020-02-13
Reworked items definition

Jaccouille:
Rewrote item definition for axes, spears, oracle, cure, omnicure and mana potion.


wurst/objects/items/Armory/AxesDefinition.wurst
```diff
@@ -5,116 +5,174 @@ import AbilityObjEditing
...
-
-    new Axe(ITEM_STONE_AXE   , Icons.bTNOrcMeleeUpOne              , ST_MODEL , "Stone Axe"   , "A", TT_COMMON , DMG_STONE , 12, asList(255, 255 , 255 ), ABIL_STONE)
-    new Axe(ITEM_IRON_AXE    , Icons.bTNOrcMeleeUpTwo              , IA_MODEL , "Iron Axe"    , "S", TT_COMMON , DMG_IRON  , 25, asList(255, 255 , 255 ), ABIL_IRON)
-    new Axe(ITEM_STEEL_AXE   , Icons.bTNSpiritWalkerAdeptTraining  , SA_MODEL , "Steel Axe"   , "Q", TT_COMMON , DMG_STEEL , 48, asList(120, 140 , 255 ), ABIL_STEEL)
-
-    new UsableAxe(ITEM_BATTLE_AXE  , Icons.bTNOrcMeleeUpThree            , BA_MODEL , "Battle Axe"  , "A", TT_BA     , DMG_BATTLE, 60, asList(255, 255 , 255 ), ABIL_BATTLE)
-    new UsableAxe(ITEM_MAGE_MASHER , Icons.bTNSpiritWalkerMasterTraining , MM_MODEL , "Mage Masher" , "A", TT_MM     , DMG_MASHER, 40, asList(255, 65  , 65  ), ABIL_MASHER)
+let TT_BA = "A two handed axe. " + TT_COMMON + " Can cast purge, removes all buffs from a target unit, and slows its movement speed." + formatCooldown(ABIL_PURGE) + formatDuration(ABIL_PURGE)
+
+ // Only purpose of this function is distinction, not sure about cooldown group, should battle axe and masher share cooldown?
```
jonathanmclaus: How do cooldown groups work for abilities on items? Does the actual ability affect shared cooldown at all?

wurst/objects/items/Armory/AxesDefinition.wurst
```diff
@@ -5,116 +5,174 @@ import AbilityObjEditing
...
+function createAxe(int axeId) returns ItemDefinition
+    return createBasicItem(axeId)
+        ..setCooldownGroup("AIas")
+
+@compiletime function createStoneAxe() returns ItemDefinition
+    return createAxe(ITEM_STONE_AXE)
+        ..setInterfaceIcon(Icons.bTNOrcMeleeUpOne)
+        ..setModelUsed(LocalItems.stoneAxe)
+        ..setName("Stone Axe")
+        ..setTooltipBasic("Trade for Stone Axe")
```
jonathanmclaus: Ahead of doing this *everywhere*, we should create a utility that updates the name and basic tooltip in this pattern (except it should say "Purchase").

wurst/objects/items/Armory/AxesDefinition.wurst
```diff
@@ -5,116 +5,174 @@ import AbilityObjEditing
...
+        ..setHotkey("A")
+        ..setTooltipExtended(TT_COMMON.format(DMG_STONE.toString()))
+        ..setScalingValue(0.7)
+        ..setLumberCost(12)
+        ..setAbilities(commaList(ABIL_STONE, "A00W"))
```
jonathanmclaus: What is `A00W`? I've surprised this even is legal because you're mixing an integer and a string in `commaList` vargs.

wurst/objects/items/Items.wurst
```diff
@@ -122,4 +77,30 @@ public class ItemChannelCast extends ChannelAbilityPreset
...
         this.setCooldown(1, cooldown)
         this.setCastRange(1, castRange)
+        this.setTargetsAllowed(1, targetsAllowed)
         this.setEditorSuffix("(Wurst)")
+
+/**
+    Use this to create an item spell who doesn't play an animation
+    Contrary to the Channel ability, WindWalk do not stop the unit current order when casted.
+**/
+public class ItemWindWalkCast extends AbilityDefinitionWindWalk
```
jonathanmclaus: Why are you using the `extends` pattern here?

wurst/objects/items/Armory/AxesDefinition.wurst
Jaccouille: Not really sure if I understood your question but here's an example :
Steel spear ability is 8 seconds cooldown
Spear ability is 2 seconds cooldown
They both belong to the same cooldownGroup
If you use steel spear, spear goes on a 8 seconds cooldown
If you use spear, steel spear goes on a 2 seconds cooldown

jonathanmclaus: If a unit had both of those abilities directly (instead of using items for them), then they would share cooldown, right? And if the two abilities share cooldown separate from the items, what does the cooldown group do for items?

wurst/objects/items/Armory/AxesDefinition.wurst
Jaccouille: This is the ability enabling the attack tree, should re-code it too.

wurst/objects/items/Armory/AxesDefinition.wurst
Jaccouille: No clue on how to do this.

jonathanmclaus: Define a new method in a local file, e.g.

    public function ItemDefinition.setNameEnhance(name)
        this
            ..setName(name)
            ..setTooltipBasic("Purchase " + name)

You could import ItemDefinition publicly to simplify imports too and then just prefer this import everywhere. Alternatively the tooltip generator could do this for you potentially, which is the better way if it is possible.

jonathanmclaus: Item's don't support tooltip generation, so you'd need to use the method approach. I had thoughts of doing something for generating kits and buildings within a single shared definition wrapper, so we can discuss this more if it's not clear.

wurst/objects/items/Armory/AxesDefinition.wurst
Jaccouille: `If a unit had both of those abilities directly (instead of using items for them), then they would share cooldown, right?`
I quickly gave steel spear & spear ability to gatherer, they don't share cooldown between them and the spears.

wurst/objects/items/Armory/AxesDefinition.wurst
Jaccouille: Cool, learnt something new.

wurst/objects/items/Items.wurst
Jaccouille: No real reason except I wrote it months ago and copy pasted it, edited.

wurst/objects/items/Armory/SpearsDefinition.wurst
```diff
@@ -73,77 +73,206 @@ let TT_RPOISON = TT_PS.format(RPOISON_DMG[0].toToolTipRed(), RPOISON_DMG[1].toTo
...
+        ..setInterfaceIcon(IRON_ICON)
+        ..setHotkey("Q")
+        ..setTooltipExtended(TT_IRON)
+        ..setLumberCost(11)
+        ..setAbilities(commaList(ABILITY_SPEAR_IRON))
+        ..setStockReplenishInterval(120)
+        ..setTintingColor1Red(200)
+        ..setTintingColor2Green(200)
+        ..setTintingColor3Blue(200)
+        ..setTooltipBasic("Trade for Iron Spear")
```
jonathanmclaus: You don't need this now.

wurst/objects/items/Items.wurst
```diff
@@ -122,4 +77,28 @@ public class ItemChannelCast extends ChannelAbilityPreset
...
+        ..setHeroAbility(false)
+        ..setItemAbility(true)
+        ..setLevels(1)
+        ..setTooltipNormal(1, "dummy spell")
+        ..setTooltipNormalExtended(1, "dummy spell")
+        ..setManaCost(1, 0)
+        ..setAnimationNames("")
+        ..setDurationHero(1, 0.01)
+        ..setCastingTime(1, 0)
+        ..setDurationNormal(1, 0.01)
```
jonathanmclaus: Does this not work at 0? And can you set the fade time higher than this if not? I worry that fully fading for any time at all will break attack / patrol orders on that unit for the enemy.

wurst/objects/items/Items.wurst
```diff
@@ -112,7 +67,7 @@ public class AttackSpeedBonusAbility extends AbilityDefinitionAttackSpeedIncreas
         this.setEditorSuffix("(Wurst)")
 
 public class ItemChannelCast extends ChannelAbilityPreset
```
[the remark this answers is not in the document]
Jaccouille: No I'll change it since you brought it up.

wurst/objects/items/Items.wurst
Jaccouille: Unit goes invisible if it's set at 0.

jonathanmclaus:
What is the effect of setting the transition time to 0? Wouldn't that cause the unique to go fully invisible immediately, breaking orders due to being briefly invisible?

Jaccouille:
> 
> 
> What is the effect of setting the transition time to 0? Wouldn't that cause the unique to go fully invisible immediately, breaking orders due to being briefly invisible?

Oh ok I get what you're asking, I suggest we used the berserk ability then.

PR #492, opened 2020-02-20
Unit restock instead of add/remove for totem

totony-misc:
$changelog: Fixed limited random stock and make totem put trolls on cooldown instead of removing

Made it so trolls don't cost lumber/gold and it's using restock mechanism instead of add/remove

Someone could prob spam click every troll, but i dont think thats a problem and anyway the totem tries to switch selection whenever u create a troll which adds a per-player timeout


totony-misc:
nvm made another PR

PR #493, opened 2020-02-20
Refactor totem selection + add random button + fix repick stock limit

totony-misc:
Seems to work, I'll close the other PR since my 2 changes affect each other

If you don't want the random change, you can cherry-pick the 4d5cf commit


wurst/objects/units/ShopObjects.wurst
```diff
@@ -0,0 +1,36 @@
+package ShopObjects
+
+import Assets
+import UnitObjEditing
+
+import LocalObjectIDs
+
+public class ShopHeroDefinition extends HeroDefinition
```
jonathanmclaus: Create a method for this, e.g.

    public UnitDefinition.setDummyStockInterface()
        this
            ..setLumberCost(0)
            ..setFoodCost(0)
            ..setGoldCost(0)
            ..setStockStartDelay(0)
            // Expect restocks to be triggered directly.
            ..setStockReplenishInterval(999) 
            ..setStockMaximum(1)

wurst/objects/units/ShopObjects.wurst
```diff
@@ -0,0 +1,36 @@
...
+      ..setFoodCost(0)
+      ..setGoldCost(0)
+      ..setStockStartDelay(0)
+      ..setStockReplenishInterval(999) // Will be programatically refilled
+      ..setStockMaximum(1)
+
+@compiletime function createDummies()
+  new ShopHeroDefinition(UNIT_RANDOM_TROLL, UnitIds.shadowhunter)
+    ..setName("Random Troll")
+    ..setTooltipBasic("Ran|c00ff00ffd|rom")
```
jonathanmclaus: Not a huge fan of this format. We don't enforce consistency right now, but in general don't highlight portions of the name. Just put the hotkey at end in its own section, e.g. `Random (D)`

wurst/objects/units/ShopObjects.wurst
```diff
@@ -0,0 +1,36 @@
...
+    ..setTooltipExtended("")
+    ..setHotkey("D")
+    ..setStockReplenishInterval(0)
+    ..setIconGameInterface("ReplaceableTextures\\CommandButtons\\BTNShoveler.blp")
```
jonathanmclaus: This is not indented properly and you should use the `Assets` package in the standard library to simplify this. In this case, that would be `setIconGameInterace(Icons.bTNShoveler)` based on [this](https://github.com/wurstscript/WurstStdlib2/blob/master/wurst/_wurst/assets/Icons.wurst).

wurst/objects/units/ShopObjects.wurst
```diff
@@ -0,0 +1,36 @@
...
+    ..setUnitSoundSet("")
+
+  new ShopHeroDefinition(UNIT_REPICK_TROLL, UnitIds.shadowhunter)
```
jonathanmclaus: Same as above - change the hotkey tooltip and the icon.

wurst/objects/units/ShopObjects.wurst
```diff
@@ -0,0 +1,36 @@
...
+
+@compiletime function createDummies()
+  new ShopHeroDefinition(UNIT_RANDOM_TROLL, UnitIds.shadowhunter)
+    ..setName("Random Troll")
+    ..setTooltipBasic("Ran|c00ff00ffd|rom")
+    ..setTooltipExtended("")
+    ..setHotkey("D")
+    ..setStockReplenishInterval(0)
+    ..setIconGameInterface("ReplaceableTextures\\CommandButtons\\BTNShoveler.blp")
+    ..setUnitSoundSet("")
```
jonathanmclaus: Does this actually matter? The unit is removed upon entry, so it's not clear to me that any sounds would actually occur.

wurst/objects/units/TrollUnitFactory.wurst
```diff
@@ -94,7 +95,7 @@ public abstract class TrollUnit
             this.normalAbilities += ","+subclassMenuID.toRawCode()
 
     function buildUnit()
-        new HeroDefinition(id, UnitIds.blademaster)
+        new ShopHeroDefinition(id, UnitIds.blademaster)
```
jonathanmclaus: Revert this.

wurst/objects/units/TrollUnitFactory.wurst
```diff
@@ -138,12 +139,6 @@ public abstract class TrollUnit
         ..setHitPointsRegenerationRate(0)
         ..setSightRadiusDay(sightRadius.a)
         ..setSightRadiusNight(sightRadius.b)
-        ..setLumberCost(3)
```
jonathanmclaus: Change this to `..setDummyStockInterface()`.

wurst/systems/core/ClassSelector.wurst
```diff
@@ -199,22 +202,19 @@ public class ClassSelector
...
+            totem.addUnitToStock(trollId, 1, 1)
+            p.removeTroll()
+            troll.remove()
 
-        owner
-        ..setGold(0)
-        ..setLumber(3)
+    function handleRandom(player p)
+        let trollId = getRandomTrollId()
+        IssueNeutralImmediateOrderById(p, totem, trollId)
```
jonathanmclaus: Does this work when the totem is a not a neutral building? I don't think it is when `-tb` is on but I could be wrong.

wurst/objects/units/ShopObjects.wurst
totony-misc: I just copy pasted the old Repick, but it didn't work because of the transparency I think, I'll fix it to be like trolls

wurst/objects/units/ShopObjects.wurst
totony-misc: The spawning sound theme was happening on my pc so i nulled it

wurst/systems/core/ClassSelector.wurst
totony-misc: ya works in my tests

totony-misc: (with -tb not sure with allied shops, I'm not even sure u can make a owner player building cause shops are considered "neutral passive")

jonathanmclaus: Not all shops are neutral. Each race in melee has a shop they can build for example. These share most of the same functionality but only share control with allies instead of all allies.

totony-misc:
> You need to manually indicate the tooltip at the end of this. We don't build any tooltips automatically right now.

not sure what you mean by that, the hotkey? I just copied the repick design and put random instead. rn there's no hotkey indication in tooltips for anything in the totem

jonathanmclaus:
We've been bad about indicating hotkeys in the past, but any new work should indicate it clearly at the end. We currently don't have a preferred format - some abilities read like `Fireball - (X)` and others read like `FIreball [X]`. The exact format doesnt' really matter, just make sure it's readable and indicated clearly at the end of the title on the tooltip.

PR #498, opened 2020-02-22
Proper name on transformation

Jaccouille:
So when you said "You can add to the library file, just don't add it to the core functionality", I understood that I could add it in transformation.wurst, let me know if I am wrong.


.gitignore
```diff
@@ -28,3 +28,5 @@ markdown_changelog.md
 
 # Python scripts
 __pycache__
+
+wbschema.json
```
jonathanmclaus: You don't need this. I added `.vscode/*` to the `.gitignore`.

wurst/lib/Transformation.wurst
```diff
@@ -115,3 +115,12 @@ public function transformUnit(unit origin, int targetID)
 
     // Destroy local resources.
     destroy key
+
+init
+    // Registering a callback to update the unit proper name after transformation
+    registerAfterEffect() (unit target, int unitID) ->
+        // Only hero have proper name
+        if IsHeroUnitId(target.getTypeId())
+            let tmpUnit = createUnit(target.getOwner(), target.getTypeId(), target.getPos(), target.getFacingAngle())
```
jonathanmclaus: Creating this at the position of the unit seems like it could have side effects. This will set off triggers for entering a region, won't it? Use the recipe island for this or some other area that is unused for actual gameplay.

jonathanmclaus: Actually `boundMax` from `MapBounds` is more sustainable. That's provided by the standard library and is used for the `InstantDummyCaster`. This way we don't have a random reference to the recipe island.

PR #505, opened 2020-02-24
Replaced Null Damage on Mage.

Jaccouille:
$changelog: Replaced Null Damage with Negative Blast as the starting ability for Mage.
$changelog: Replaced Negative Blast with Spirit Prison within the spellbook for Mage.

So I made this skillshot which snare first target hit

I also updated ItemWarp since I was at it because I noticed that I forgot to update the Z coordinate, missile would disappear when going through cliff/high ground


wurst/objects/abilities/mage/unsub/Wabbajack.wurst
```diff
@@ -0,0 +1,168 @@
...
+
+    function onImpact()
+        this.exploded = true
+        InstantDummyCaster.castTarget(this.ownerUnit.getOwner(), ABILITY_WABBAJACK_CAGE, 1, Orders.ensnare, this.target)
+
+    function spawnTemporaryVision(real radius, real duration)
+        let vis = createVision(ownerUnit.getOwner(), this.pos.toVec2(), radius, true)
+        vis.start()
+        doAfter(duration) ->
+            vis.destr()
```
jonathanmclaus: This feels weird if the unit is released early, either by dying or being cured (assuming that removes the effect). You could use a fake faerie ability without armor reduction or an art effect?

Jaccouille: True.

Jaccouille:
> 
> 
> Do you have strong feelings about the name? I'd rather not introduce more references into the game.

No, I randomly named it wabbajack because I remember you suggesting it has a replacement for metronome.

wurst/objects/abilities/mage/unsub/Wabbajack.wurst
Jaccouille: However there is no field to set a vision range with faerie fire, it uses the target vision range, I don't really want the vision range to be too big and give for example, vision of the target base. I just want the spell caster to know that he hit a target, the effect last 1.75 sec on hero so I think that's negligeable.

jonathanmclaus:
> > Do you have strong feelings about the name? I'd rather not introduce more references into the game.
> 
> No I don't, I randomly named it wabbajack because I remember you suggesting it has a replacement for metronome.

That was slightly more thematic and it was replacing another reference (Metronome is taken from Pokemon). What about just calling it "Torment Soul"?

wurst/objects/abilities/mage/unsub/Wabbajack.wurst
jonathanmclaus: OK.

Issue #482, opened 2020-02-15
Troll getting slowed indefinitely


Jaccouille:
I played a game, two of my teammate went into a fight, they went into a fight including unsub gatherer, mage and thief. When they came back, they had the slow debuff, the one given by negative blast/metronome, I didn't watch the fight and forgot to save replay, they got slowed for the whole game.

My guess is on gatherer armor salve, it's based off frost armor so I bet it can give the debuff, will check it out.

jonathanmclaus:
Fixed by #509

Issue #475, opened 2020-02-10
Killing hydra triggers an error


Jaccouille:
![](https://i.gyazo.com/c4399a6dab21330213e86377acb8592c.jpg)

jonathanmclaus:
Fixed by #509

jonathanmclaus:
![image](https://user-images.githubusercontent.com/5731591/79348519-17413400-7f03-11ea-8397-759157b55ac2.png)

Still broken as of 3.6

Issue #435, opened 2019-12-13
Sleep protection does not protect stats


jonathanmclaus:
Enabling sleep protection blocks allies from casting sleep on you but the order still causes you to lose stats.

jonathanmclaus:
Fixed by #509

PR #509, opened 2020-02-25
Fixed miscellaneous bugs.

Jaccouille:
$changelog: Gatherer armor salve no longer permanently slow enemies

I don't know what the hell is happening with the git history, but I think wurst setup add a new line on first line of build file when you update the project.
I don't know why I get "Nerfed living clay" commit in my history.

I fixed armor salve issue #482, it didn't have the base frost armor spell as baseId but a custom one, the armor salve spell isn't created at compile time due to that, hence why the tooltip shows "0 bonus armor". For some reason, the item still casted a frost armor buff effect, and the slow would never go off, funny behavior.

As for the Hydra #475, I am not sure but I believe onLesserHydraDeath destroys all hydra instances, so the entity instance doesn't exist anymore yet it tries to call removeUnit(), hence the error message. I am not at ease with unitEntity so I just moved up the removeUnit() call, didn't see any issue in game.

For the sleep issue #435, well I rewrote the sleep abilities defintion and re-arranged the trigger effect, I believe the code is clear


wurst/objects/abilities/Sleep.wurst
```diff
@@ -11,30 +12,85 @@ import ChatUtils
...
+let MANA_RESTORED_BY_HUTS = 200.
+
+let SLEEP_OUTSIDE_HP_PENALTY = 20.
+let SLEEP_OUTSIDE_HEAT_PENALTY = 15
+
+let INVUL_DURATION = 1.0
+let COOLDOWN = 10.
+let DURATION_HERO = 6.0
+let CAST_RANGE = 300.
+let TARGET_ALLOWED = "friend,ground,hero,self"
```
jonathanmclaus: Can you use the predefined `TargetTypes` as a comma list from assets?

PR #515, opened 2020-02-26
TRIAL : Changed gatherer base spell for item limited version of telegather

Mephistophelesss:
$changelog: Gatherer base ability has been changed and can now telegather 5 items.

### DIFF
--- a/wurst/objects/units/TrollUnitTextConstant.wurst
+++ b/wurst/objects/units/TrollUnitTextConstant.wurst
@@ -97,7 +97,7 @@ public constant  NORMAL_SPELLS_PRIEST       = commaList(ABILITY_THE_GLOW)
 public constant  NORMAL_SPELLS_BEAST_MASTER = commaList(ABILITY_PET_SPELLS, ABILITY_SPIRIT_BEAST )
 public constant  NORMAL_SPELLS_THIEF        = commaList(ABILITY_TELEPORT)
 public constant  NORMAL_SPELLS_SCOUT        = commaList(ABILITY_PING_ENEMY)
-public constant  NORMAL_SPELLS_GATHERER     = commaList(ABILITY_SALVE_RECIPE)
+public constant  NORMAL_SPELLS_GATHERER     = commaList(ABILITY_TELE_GATHERER)
 
 // Sub Hero Spells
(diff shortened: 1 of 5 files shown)


jonathanmclaus:
Where is `ItemRecipe` defined?

jonathanmclaus:
Nevermind, forgot that was already included in `wurstExtLibs`.

Mephistophelesss:
you're right but let's move to #516 (this merge was meant for telegathering my bad)

wurst/objects/units/TrollUnitTextConstant.wurst
```diff
@@ -97,7 +97,7 @@ public constant  NORMAL_SPELLS_PRIEST       = commaList(ABILITY_THE_GLOW)
 public constant  NORMAL_SPELLS_BEAST_MASTER = commaList(ABILITY_PET_SPELLS, ABILITY_SPIRIT_BEAST )
 public constant  NORMAL_SPELLS_THIEF        = commaList(ABILITY_TELEPORT)
 public constant  NORMAL_SPELLS_SCOUT        = commaList(ABILITY_PING_ENEMY)
-public constant  NORMAL_SPELLS_GATHERER     = commaList(ABILITY_SALVE_RECIPE)
+public constant  NORMAL_SPELLS_GATHERER     = commaList(ABILITY_TELE_GATHERER)
```
jonathanmclaus: Really dumb question, but does Gatherer not have the ability to ping all items anymore? Also does he not have room for 2 normal abilities? His second row could be Salves, Radars, Telegather, Skill Menu I think.

Mephistophelesss: Yeah the ping ability is now leveled remember, it also reveals the item on the groud. I also though it could have a 10% to telegather back to fire with telegathering enabled (or something similar). Also every sub class have 2 available slots for ability (one of the base ability and the other for leveling one). I dont think we have space for salve recipe but we can add a level to the spellbook (with the recipe in it) and let gatherer have it by default. (Should still be able to get level 2 3 4 spellbook by leveling)

Mephistophelesss:
Can we close this?

Jaccouille:
I actually like the idea, tele gathering is a pretty original ability, I've never seen a similar concept in another game, I find it entertaining and I would like to make it accessible to new player so they enjoy it right away.

Also I hate gatherer salve.

PR #527, opened 2020-03-07
Changed the UI for the status bar.

Mephistophelesss:
$changelog: Changed the UI to display effects on units more prominently.

Preview:

image

### DIFF
--- a/wurst/systems/ui/Status.wurst
+++ b/wurst/systems/ui/Status.wurst
@@ -7,10 +7,11 @@ function moveStatusFrame()
 
     // Move the the buff bar to above the unit panel.
     getOriginFrame(ORIGIN_FRAME_UNIT_PANEL_BUFF_BAR)
-        ..setScale(2)
         ..clearAllPoints()
         ..setAbsPoint(FRAMEPOINT_LEFT, 0.22, 0.16)
+        // Enlarge it.
+        ..setScale(2)
 
 init
-    // Execute this for all players immediately.
+    // Modify the status bar for all players immediately.
     moveStatusFrame()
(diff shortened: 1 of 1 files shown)


wurst/systems/ui/Status.wurst
```diff
@@ -0,0 +1,18 @@
...
+    // Assign the origin frame of the buff bar to a variable
+    let buffs = getOriginFrame(ORIGIN_FRAME_UNIT_PANEL_BUFF_BAR)
+    // Assign the origin frame label of the buff bar to a variable
+    let bufflabel = getOriginFrame(ORIGIN_FRAME_UNIT_PANEL_BUFF_BAR_LABEL)
+
+    // Clear the Points that link the label with the bar
+    bufflabel.clearAllPoints()
+
+    // We increase the scale and set the point to be just above the portrait
+    buffs.setScale(2)
```
jonathanmclaus: Can you use the `..` notation to condense calls in this file? E.g.

    buffs
        ..setScale(2)
        ..clearAllPoints()
        ..setAbsPoint(FRAMEPOINT_LEFT, 0.22, 0.16)

Mephistophelesss:
This is a pretty nice feature, but it requires more testing, I don't know if it can bug with different monitor size. If some of you guys can give it a try before we merge, I'd be thankful.

jonathanmclaus:
It looks mostly fine for me - it seems to cut a bit into the frame for the bottom UI (just the artwork).

image

I am worried that players will be annoyed by it blocking clicks on part of the screen they used to be able to see - for example the effect can cover up an item or a unit when you're trying to target it.

Issue #202, opened 2019-01-04
Unroll experience


jonathanmclaus:
Currently, each level gets a new experience reduction. This could be fixed by unrolling the experience per level and then the only factors that would need to reduce the experience to match the current system is 0.75 for sub-classes and 0.5 for super sub-classes.

This could potentially affect how the Medallion of Courage is used, as that cuts off several levels and affects ability score changes.

This would reduce the efficacy of Magic granting +500 experience, which is currently the only source of unmodified experience gain.

tsclaus:
Did it ever get figured out whether the reduction can be applied to half a
kill? E.g. does killing a bear then an elk result in the same experience as
killing an elk then a bear? In the past, the kill order has certainly
mattered with the mammoth as being level 2 when you solo it results in less
XP than soloing it at level 1.

theQuazz:
@tsclaus I fixed that, currently kill order does not matter

Jaccouille:
Is this issue still relevant?

jonathanmclaus:
Yes, I think so.

theQuazz:
Yes it is

PR #533, opened 2020-03-11
Added Score Board

Jaccouille:
$changelog: Added -ps (player-scoring) mode, display a score board to observer & player who are victorious/defeated


jonathanmclaus:
Why is this behind a mode?

Jaccouille:
> 
> 
> Why is this behind a mode?

In case there's any issue (code causing desync) I want to be able to disabled it so player can still play the version.

wurst/systems/boards/TribeBoard.wurst
```diff
@@ -131,8 +189,25 @@ init
...
+            TribeBoard
+                .findForPlayer(owner)
+                .setTrollLevel(owner)
+
+    registerAfterEffect() (unit target, int unitID) ->
+        if target.isTroll()
+            let owner = target.getOwner()
+            // For some reason, the troll level isn't updated in time with this event, but the name is
+            // waiting a bit to get the correct level
+            nullTimer() ->
```
jonathanmclaus: Does this not work with the null timer? In that case the transformation library should probably include the null timer.

Jaccouille: It works with the nulltimer.

jonathanmclaus: Oops, had a typo. I meant: does it work without the `nullTimer`? If you don't have the `nullTimer`, are values for the original troll used?

jonathanmclaus: Oh, nevermind. This is because the level is reset not by the transformation system. Right now it happens:

1. Transformation system changes the unit type and that's it.
2. It fires callbacks, including this.
3. The `TrollUpgrade.wurst` package changes the level post transformation.

We should either have a callback system for troll transformation specifically, move the level reset prior to the transformation, or switch to a single level track.

jonathanmclaus:
Holding off on this until #202 is addressed so that the transformation hook works better.

PR #536, opened 2020-03-15
Improved healing reduction.

jonathanmclaus:
$changelog: Fixed bug where healing tags could be seen through fog.
$changelog: Reduced effects of healing reduction.
$changelog: Added more information regarding healing reduction in quests and resources.


jonathanmclaus:
Changelog comments are included in the squash commit for #537 because I forgot to include them in this squash commit.

PR #541, opened 2020-03-21
Fixed Iron Shield recipe.

jonathanmclaus:
$changelog: Fixed bug preventing Iron Shield from being crafted.


jonathanmclaus:
Erased the squash commit manually because I forgot to include the changelog message and redid this in #542

PR #550, opened 2020-07-25
Druid adjustement

Jaccouille:
Formatted a few tooltips to make it more readable.


wurst/objects/abilities/beastMaster/Druid/Sentinel.wurst
```diff
@@ -19,8 +19,8 @@ let BUTTON_POS_X = 0
 let BUTTON_POS_Y = 0
 let HOTKEY = "Q"
 let NAME = "Sentinel"
-let TOOLTIP_BASIC = "Send out a Sentinel Owl"
-let TOOLTIP_EXT   = "Send out a Sentinel Owl to watch from a nearby tree."
+let TOOLTIP_BASIC = "Sentinel Owl"
+let TOOLTIP_EXT   = "Send out a Sentinel Owl to watch from a nearby tree. The Owl will disappear if the tree is attacked."
```
jonathanmclaus, line 23: I thought the tree has to be fully cut down, not just attacked. If that is the case, then the tooltip should make that clearer.

HallerJohn: I can't get my wurst working rn to verify. But iirc all it takes is one hit to send the owl away.

Jaccouille: Just a basic attack is enough to make it disappear.

Issue #553, opened 2020-07-27
Icon for pet revival does not follow standards


jonathanmclaus:
The reincarnation icon for pets does not have the passive background, so it may confused players who assume it is castable.

jonathanmclaus:
Closed by #555

PR #555, opened 2020-07-27
Improved UI elements for pet reincarnation.

Jaccouille:
Issue #553 

url


jonathanmclaus:
Can you update the tooltip to include the fact that it revives next to you, e.g.

    The pet will revive next to its master 90 seconds after it dies.

Issue #554, opened 2020-07-27
Elk (and possibly other pets) cannot damage Mammoth


jonathanmclaus:
The elk pet can attack the mammoth but will do no damage, regardless of how much base damage the pet has.

Jaccouille:
This is due to the` Combat - Damage Bonus Table - Pierce`, it has been changed to 0% against medium armor, Elk deal pierce damage and mammoth got medium armor.

![url](https://i.gyazo.com/c2937b40366f9310d48d4198b5a1f591.png)

jonathanmclaus:
This is problem a "fix" for preventing Mammoth being killed by ensnare traps and towers. I don't think we need to worry about that anymore as long as there's no way to trap and glitch its pathing.

jonathanmclaus:
Fixed by #561

PR #561, opened 2020-08-04
Made minor improvements.

Jaccouille:
$changelog: Fixed bug where Blur would cause lag spikes.
$changelog: Allows attacks with piercing damage to hurt bosses.

Pretty sure blur didn't cause lag spike after the last fix we made months ago so I didn't bothered optimizing this target check, now it causes huge lag with ~3 units, added an additional condition and didn't notice more lag afterward.

Arbitrarily changed pierce damage bonus from 0 to 0.5 following #554 just so we can push some discussion on it.


jonathanmclaus:
Does this mean that targets that are dazed earlier might have the effect end sooner than targets dazed later, even if both stand next to the Escape Artist?

Consider the following:
1. The Escape Artist stands still while using Blur.
2. Unit A approaches the Escape Artist, stands next to him for 2 seconds, and then moves away. He will be Dazzled for 1 more second.
3. Unit B does the same, but leaves after 3 seconds. He is Dazzled for another 3 seconds.

That doesn't seem fair.

Jaccouille:
This statement is correct and I didn't think too much about fixing in a "cleaner" way, I believe this isn't much of an issue to be honest, but well, I never play Thief.

jonathanmclaus:
During each iteration, you could track which units are viable targets for dazzle, but avoid casting it on them if they already have it. During that iteration, prior to updating this group, check if any existing members are no longer viable targets. For those units, recast the spell. Adding and removing units from this group should be significantly faster than casting the dazzle ability.

This minimizes casting the dazzle ability but still prevents that. Maybe this is spending too much time on a problem unlikely to be a balance concern though.

PR #562, opened 2020-08-04
Added a debuff on Anger's target after the effect ends.

Jaccouille:
$changelog: Added an Exhaust debuff that is applied to the target of Anger after its effect ends, which reduces its attack speed by 50% and movement speed by 25% for 4 seconds.

Some people reported anger being too strong, I think this debuff makes sense, can tweak the numbers.


jonathanmclaus:
$changelog: Added an Exhaust debuff that is applied to the target of Anger after its effect ends, which reduces its attack speed by 50% and movement speed by 25% for 4 seconds.

PR #566, opened 2020-08-05
Added area of effect indicator for wards to match sight radius.

Jaccouille:
$changelog: Added area of effect indicator for abilities that place wards which matches the sight radius of the ward.

A quality of life suggestion.

### DIFF
--- a/wurst/objects/abilities/scout/WardArea.wurst
+++ b/wurst/objects/abilities/scout/WardArea.wurst
@@ -12,6 +12,7 @@ import ColorUtils
 import LocalObjectIDs
 import ToolTipsUtils
 import StringExtensions
+import LivingClay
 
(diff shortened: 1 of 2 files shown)


jonathanmclaus:
Is this the normal target image? I don't remember seeing the axe in the middle before. I thought there was one with radial symmetry.

Jaccouille:
Initially I thought it depended on the race you selected in the lobby or the Race specified on the ability object, but I didn't manage to get any other target image than this one, supposed to be the Orc.

jonathanmclaus:
I worry that this one isn't that precise on where the ward actually is. I'll check out how Inferno or Blizzard do it.

jonathanmclaus:
Turns out this is based on the caster's race. More information [here](https://www.hiveworkshop.com/threads/how-to-get-a-different-targeting-image.307690/). Will approve this and we can update via imports later if we want.

PR #572, opened 2020-08-20
Wurstified buildings kits.

Jaccouille:
Also changed **omnidefender** variable name to **omnitower** to avoid confusion.


jonathanmclaus:
This is a step in the right direction but we should also just automate this in the long run.

PR #577, opened 2020-08-22
Fixed bugs; balanced various features.

Jaccouille:
$changelog: Fixed bug where Troll Battle Call would cause mana to regenerate.
$changelog: Reduced Druid base speed from 330 to 300.
$changelog: Increased lifespan of bushes created by Breath of Life bush from 30 to 60 seconds.
$changelog: Increased level requirements for Dementia Master and Sage from 7 to 8.
$changelog: Reduced Adult jungle bear base damage from 33 to 21.
$changelog: Changed Tribe Gold and Tribe Buildings modes to be enabled by default.
$changelog: Changed EMP to disable sleeping at buildings in its area of effect.
$changelog: Added target effect visible when fires burn nearby buildings.

Troll Battle Call shouldn't give mana regeneration, it was either a mistake on my part when I recoded it, or old spell tooltip didn't mention the mana regeneration.

Druid had same base movement speed as Juggernaut, was way too fast.

I asked Quazz about the "removeUnit" attribute in UnitEntity, I don't understand why it is set as private when we should be able to alter it following our needs with the inherited units. I set it as public and "false" for Buildings Entities so they now have a proper death animation. Fixed mixing pot death animation and replaced the crystal shard model with a pretty purple herb

Breath of life seems underwhelming, increased duration so you can have atleast 2 set of bushes up at the same time, might want to increase it again.

Dementia Master & Sage have been reported as being too strong/too easy to acquire, increased level requirement

Adult Jungle Bear base damage is simply too high, feedback says he is was too strong.

TG & TB mode have been requested to be set as default mode.

EMP is supposed to disable a base defense, I believe we can consider sleeping building as "defense" as the sleep cast they provide is a huge advantage when your team is composed of spell caster, might make EMP useful again.

I also wurstified the camp fire/heat related abilities and added a FX on building being burned by Camp/Mage fire, thought it help people understand that their buildings are being burned.


jonathanmclaus:
What is the change in the .mdx?

PR #587, opened 2020-08-30
Fixed bugs associated with Spirit Prison.

Jaccouille:
$changelog: Fixed bugs where Spirit Prison could hit multiple targets and be used multiple times without cooldown.

Spirit had a conflicting baseOrderId with quick make mixing pot.
I don't really understand how am I supposed to use the order banning, but I figured "blink" was banned due to thief teleport, so I set spirit prison baseOrderId to "blink" for now.


jonathanmclaus:
Yeah, it's a good way to handle order IDs for now - anything that another class uses for a real ability should be good to go, assuming you don't do that multiple times for the same class on accident, e.g. can't use `blink` for anything else on Mage now and we might forget about that because it's not immediately clear which order IDs Mage uses.

PR #585, opened 2020-08-29
Cleaned up old object definitions.

Jaccouille:
Merged #581 prior to that

I went over all old object definition in units/abilities/raw/buffs and rewrote them in readable wurst code.

I added a python script to format the old wurstOutput.txt definition we had in readable wurst, don't know if it's worth pushing but it saved me a lot of time, I just copy/pasted the readable output and fixed a few mistake as the script isn't perfect (a few formatting mistake with abilities & buff definition)

I removed every object definition that wasn't actually used in the map at this moment such as disease meat, smoke meat, old elementalist mode & its spell,

I reset the ability & buffs modified builtin.
I didn't touch gatherer salve as I don't know if we're going to keep them.
I didn't touch flying pet abilities, but I plan to remove them in the hatchery rework.
I didn't touch rendo abilities as I plan to rework them.

PR #588, opened 2020-08-30
Reenabled custom troll skins.

Jaccouille:
So I read #546 and the comments left being in Skins.wurst, not sure if I understood since I didn't run into the issue, is it possible/reasonable to re-enable skins?

Some people (Mambo) wants their skins back, it was disabled in the first place because I was lazy and didn't rewrote their unit definition, plus I kind of didn't like this skin idea, most of the time, they weren't used and were the heaviest custom models file contained in ITT.

But I guess things such as "ranking" & "reward" motivate people to play, plus it would be a way to show gratitude to people still playing the game, reporting bugs and streaming/uploading replays.


jonathanmclaus:
Can we leverage the new `BlzCreateUnitWithSkin` instead? I think part of the problem with skins is that it's easy to write code to check the type of unit and then miss that comparison if the ID isn't the normal one.

Jaccouille:
> 
> 
> Can we leverage the new `BlzCreateUnitWithSkin` instead? I think part of the problem with skins is that it's easy to write code to check the type of unit and then miss that comparison if the ID isn't the normal one.

Well to my surprise, it actually work, you just need to create another unit with a different model and pass this unit Id as a skin.
It swap the model and scaling value, but keep the original unit Id and its abilities.

Jaccouille:
What is the purpose of this [check](https://github.com/island-troll-tribes/island-troll-tribes/blob/3d591a0ca019c07cf5148d9c2724d3a1725ceacf/wurst/systems/commands/Skins.wurst#L109)? Upon subbing, onEnter is called twice upon subbing.

jonathanmclaus:
You linked a check outside of `onEnterHandler`? Anyway, that check makes sure that you can use a skin after a transformation, but only if the transformation is one used for changing class. So if you had multiple skins for Wolf Form (for example), you can't switch skins after switching between forms on BM.

As it's coded now, you can only apply a skin when you first pick a troll and after changing class, meaning that we can't add skins for secondary Shapeshifter models. We could either change to allow you to pick a skin the first time a unit assumes a given model, or just every time. I don't think it's too abusable now. It might fuck up targeting though.

Jaccouille:
Ok, well, upon subbing, we enter in both closure & go into `onEnterHandler`, which make the skin available message display twice.

jonathanmclaus:
`onEnter` shouldn't be fired when you sub. The unit is still the same unit.

Jaccouille:
In this screenshot, I put a print(target.getName()) just before the skin message display

jonathanmclaus:
Duplicate message was actually caused by a transformation after effect creating a duplicate unit, required to repick a new proper name for the hero.

wurst/systems/commands/Skins.wurst
```diff
@@ -86,8 +85,14 @@ function onEnterHandler(unit target)
...
     registerCommand(target.getOwner(), skin.featureName) (owner, args) ->
         // Transform the unit to impart the custom skin.
-        // TODO: Generate skins to match trolls and uncomment this.
-        // transformUnit(target, skin.newUnitTypeId)
+        BlzSetUnitSkin(target, skin.newUnitTypeId)
+
+        // For some reason the hero glow disappear on skin switch
+        // but the unit still has the ability, resetting it to "re-enable" the glow
+        // Master Healer skin already has hero glow
+        if skin.newUnitTypeId != UNIT_MASTER_HEALER_1
```
jonathanmclaus, line 93: Can you check for whether the unit has the ability directly, e.g. `if target.hasAbility(ABILITY_HERO_GLOW)`? Checking type ID directly is fragile.

wurst/systems/commands/Skins.wurst
```diff
@@ -58,12 +58,12 @@ class CustomSkin
...
-    new CustomSkin(UNIT_MAGE,          UNIT_MAGE_1,          "skin-mage-1"   )
-    new CustomSkin(UNIT_MASTER_HEALER, UNIT_MASTER_HEALER_1, "skin-mh-1"     )
-    new CustomSkin(UNIT_BOOSTER,       UNIT_BOOSTER_1,       "skin-booster-1")
-    new CustomSkin(UNIT_ELEMENTALIST,  UNIT_ELEMENTALIST_1,  "skin-ele-1"    )
+    new CustomSkin(UNIT_BEAST_MASTER,      UNIT_BEAST_MASTER_1,  "skin-bm-1"     )
+    new CustomSkin(UNIT_HUNTER,            UNIT_HUNTER_1,        "skin-hunter-1" )
+    new CustomSkin(UNIT_MAGE,              UNIT_MAGE_1,          "skin-mage-1"   )
+    new CustomSkin(UNIT_MASTER_HEALER,     UNIT_MASTER_HEALER_1, "skin-mh-1"     )
+    new CustomSkin(UNIT_BOOSTER,           UNIT_BOOSTER_1,       "skin-booster-1")
+    new CustomSkin(UNIT_ELEMENTALIST_NEW,  UNIT_ELEMENTALIST_1,  "skin-ele-1"    )
```
jonathanmclaus, line 66: Will this get messed up by your other PR? I thought we were removing this variable.

Jaccouille: This variable is indeed removed in #585 , I'll check it out once this is merged.

jonathanmclaus (approved):

PR #589, opened 2020-08-30
Fixed Elk attack animation.

Jaccouille:
$changelog: Changed Elk attack to use a melee animation.

Original Stag critter model has 2 attack animation but for some reason they do not play as intended and the unit get stuck in the stand ready animation.

I download a model from  [HiveWorkshop](https://www.hiveworkshop.com/threads/blizzards-bugged-resources.237708/page-12#post-2814308) which has its attack animation fixed.


jonathanmclaus:
Attack animation looks good, but there's no need for a missile now, right? I was planning on removing that once it actually uses its head to poke people.

PR #595, opened 2020-09-06
Reworked supersub for Beastmaster.

Jaccouille:
$changelog: Renamed Ultimate Form as Jungle Tyrant.
$changelog: Replaced Ultimate Form's auras with Devour, which grants various abilities based on usage.
$changelog: Reduced Ultimate Form base movement from 390 to 330.
$changelog: Reduced Ultimate Form base attack speed from 1.40 to 1.55.

Well Rendo with panic still runs at top speed since he has high stat, but I think slows will be more efficient if we reduce his base movement speed.

Devour regenerate corpses * 25 health point, I didn't add this in the tooltips because I couldn't find a good way to express it in an english that would fit the tooltip.

Hawk Eye : Gains true sight ability for the duration (25 seconds)

Snake Toxin : Add poison slow to Rendo basic attack, the slow stack with poison spear/thistle, slow attack speed by 30%, movement by 20%, poison last 5 seconds, the ability on Rendo last 15 seconds

Panther Instinct : Deny any damage received for 4 seconds.

Elk Jump : Just a jump, short range though, near twice less than escape artist jump

Wolf Bite : Add a debuff on target, target takes damage when moving, the more ground covered, the higher the damage, maximum damage's around ~200. Leave a blood trail every 2.4 seconds.

Once an ability is used, it becomes locked and you need to devour the animal to unlock again. The abilities got cooldown, this way you cannot continuously jump around and devour elk, well you'll have to wait 60 seconds atleast.

Rendo's evolution is a hero spellbook containing following spell :
Level - 1 : Hawk Eye, Snake Toxin
Level - 2 : Panther Instinct, Elk Jump
Level - 4 : Wolf Bite

I also reworked the spellbook object definition since it wasn't a flexible design. There was an issue with spellbook sharing abilities, like rendo's aura, it was due to Rendo the same baseOrderId for 2 spellbook.


jonathanmclaus:
What do you think of renaming the class with this change? Ultimate Form doesn't really mean anything and is redundant with Super Sub. I think "Predator" is a bit too simple but is at least captures the idea behind the class better. I think Sasquatch is also accurate but a bit silly.

Jaccouille:
Oh yes, I wanted to rename it too, so "ultimate form" always sounded to me like, he is a the top of the food chain, apex predator of the jungle etc... also I think ITT has pretty original name for classes such as "dementia master" or even "Sage", that's a term I do not see very often in english language, wanted to find a somewhat original expression for Rendo.

I thought `Jungle Tyrant` sounded cool/oppresive, like he has pet but they obey him out of fear, I think tyrant's powerful word.
Warcraft 3 actually named this model `Jungle Stalker` but that's less cool.

I came upon this [skin](https://www.hiveworkshop.com/threads/junglebeast-blp.253276/#resource-41831), made me think of the word "tyrant"

wurst/utils/TextTagUtils.wurst
```diff
@@ -0,0 +1,35 @@
...
+
+
+/**
+    This function create a textTag upon the target unit and make it fade away upward
+    Visible only by allied player
+**/
+public function createTeamFadingTextTag(unit target, string text, color color) returns texttag
+    // Create a text tag to visualize the effect.
+    if localPlayer.canSee(target)
+        return CreateTextTag()
```
jonathanmclaus, line 14: Why don't you forward the call to the below function? This seems like unnecessarily duplicated code unless you expect the behavior like velocity to not always be the same (which seems doubtful).

jonathanmclaus (changes requested):

wurst/objects/abilities/JumpSystem.wurst
```diff
@@ -113,10 +160,16 @@ class JumpInstance
...
+            flashEffect(Abilities.crushingWaveDamage, targetPos)
         if u.getTypeId() == UNIT_ONE
             if InstantDummyCaster.castImmediate(u.getOwner(), JUMP_STOMP_ID, 1, OrderIds.creepthunderclap, u.getPos()) == false
                 error("Jump Stomp cast failed")
             flashEffect(Abilities.warStompCaster, u.getPos())
 
     ondestroy
         finishJump()
+
+init
```
jonathanmclaus, line 175: Why have an empty `init` block?

wurst/objects/abilities/JumpSystem.wurst
```diff
@@ -9,33 +9,80 @@ import InstantDummyCaster
...
         else
             printTimedToPlayer("Requires test mode!", 5., triggerPlayer)
 
-public function performJump(unit u, vec2 targetPos)
+function canJumpToTarget(vec2 targetPos) returns bool
+    return targetPos.isTerrainWalkable()
+
+public function onStartCast(real maxRange)
+    let spellId = GetSpellAbilityId()
+    if spellId != ABILITY_JUMP and spellId != ABILITY_RENDO_ELK_JUMP
```
jonathanmclaus, line 40: Why does this need to be checked? What other places can call this function?

jonathanmclaus (changes requested):

wurst/objects/abilities/JumpSystem.wurst
Jaccouille: Forgot it.

wurst/objects/abilities/JumpSystem.wurst
Jaccouille: Only EA & Rendo jump call this function, this part is indeed not required anymore.

wurst/objects/abilities/beastMaster/UltimateForm/Devour.wurst
```diff
@@ -0,0 +1,124 @@
...
+        let entity = UnitEntity.findForUnit(target)
+        // Check if unit eaten is an hostile entity
+        if entity instanceof Hostile
+            let hostile = entity castTo Hostile
+
+            // Disable item & corpses drop
+            hostile.setDropOndeath(false)
+
+            // Dealing 50 pure damage to kill the target
+            UnitDamageTarget(caster, target, HEALTH_THRESHOLD.toReal(), false, false, ATTACK_TYPE_CHAOS, DAMAGE_TYPE_UNIVERSAL, null)
```
jonathanmclaus, line 93: Is this guaranteed to kill the target? Can armor types be configured to resist Chaos damage, in which case we might forget about this?

Jaccouille: Your statement is right, I chose Chaos damage because it is designed to 100% damage on every kind of armor.

Jaccouille: Could just dealt something like 99999 damage, that's what they do in League of Legend when they want to make sure a target is executed.

jonathanmclaus: Yeah, that would work except that I think if an armor type reduced damage by 100% (like Mammoth did for piercing), then that still wouldn't kill the unit. I think this is fine for now but the experience library we have should just directly support killing a unit and awarding experience for a given killer without having to go through damage indirectly.

I think that the 99999 could also throw off the scoreboard, so lets keep it as is for now.

PR #596, opened 2020-09-07
Updated scoreboard logic.

Jaccouille:
$changelog: Changed scoring to be enabled by default, keeping the -ps mode to disable it.
$changelog: Fixed bug where healing done by Cloak of Healing was not counted for scoring.
$changelog: Changed scoring to not count wasted healing when exceeding maximum health.
$changelog: Updated scoring to differentiate healing done on self and allies.

I set player score as default mode because I think it is smarter, people might just forget to enable it, plus I've gained confidence after a few testing game


wurst/systems/boards/PlayerScore.wurst
```diff
@@ -275,19 +287,35 @@ init
...
+                            playerScores.get(healerOwner.getTribe().getNonMember()).addAllyHealing(healingAmount.toInt())
+
+                    // Do not include self healing consumable items
+                    // I have mixed opinion on this, we should differentiate healing done from ability/item
+                    // This we can "quantify" how much cooked meat priest can actually save through healing spells
+                    // Still, I think having the healing done from meat could also be useful
+                    else if healerOwner == target.getOwner() and instance.getHealingType() != HealingType.ITEM
+                        playerScores.get(healerOwner).addSelfHealing(healingAmount.toInt())
+
+                        if gameConfig.isTribeOwnsBuildingsEnabled()
```
jonathanmclaus, line 317: What is the point of this line? Just so that the tribe is equal to the sum of the members instead of `0`? Why is that necessary?

wurst/systems/boards/PlayerScore.wurst
```diff
@@ -94,8 +95,8 @@ public class PlayerScore
             goldAmount += goldAdded
 
             // Updating gold amount for tribe player
-            // TODO: Disable if -tb isn't on?
-            playerScores.get(this._player.getTribe().getNonMember()).goldAmount += goldAdded
+            if this._player.getTribe() != null
```
jonathanmclaus, line 98: Same question as below. Is this just enforcing that the total for a tribe is the sum of the members? That should be computed retroactively if anything.

jonathanmclaus (changes requested):
How does this change the behavior for Cloak of Healing like the changelog mentions? It looks both types of healing worked before the change.

wurst/systems/boards/PlayerScore.wurst
Jaccouille: > 
> 
> How does this change the behavior for Cloak of Healing like the changelog mentions? It looks both types of healing worked before the change.

Cloak Of Healing healing instances do not register the caster, I simply didn't put the caster as parameter of the constructor, when no caster is specified, the instance [set the caster as target.](https://github.com/island-troll-tribes/island-troll-tribes/blob/ffde7d8cf80fd3df5dc807eea06919193635c24a/wurst/lib/HealingSystem.wurst#L44)

My condition to detect "healing done to allies" was `if caster != target`, I could have fixed healing cloak healing by specifying the caster in the constructor, however I also needed to handle healing casted by dummyCaster, e.g, sage light gate, so I changed the condition to detect the healing instance owner instead.

Note: OK I wrote this to explain how I fixed it, just to realize, that I actually didn't fix it, healing cloak instances need the caster unit, otherwise they will set the caster as target and healingOwner will equal healingOwner, shaking my head.

Jaccouille: > 
> 
> What is the point of this line? Just so that the tribe is equal to the sum of the members instead of `0`? Why is that necessary?

playerScores lookup the instances using the player variable, I assumed the player score would work on par with tb, as tb assign a playerId to the tribe, I could consider the tribe as a player and create a player score instance for it. 

If tb is disabled, there's no row displayed for tribe computed score.

The issue here, is that I do not create a player score tribe if tb is disabled, yet I try to access it through playerScores IterableMap, which result in a nullPointerException, big mistake.

Adding those if conditions prevent the "illegal access" if that's the right expression, but I can seem that the design is flawed.

Jaccouille:
> You should create columns (but maybe don't call them that because it implies orientation), essentially a value per player. Then create a scoreboard from a list of instances of that type. And modify the instances directly or look them up from the scoreboard, but at least don't write code per value more than you have to. A method name is going to lead a lot of kludge very quickly

By method name, are you referring to the scoreboard method such as ` scoreBoard.board.getItem(0, 0)` or playerScore method `function addDamage(int damage)`?

jonathanmclaus:
I mean something like this.

```
let damageValues = new ScoreboardValue(constructor args like picture for the column)  // In file that tracks damage for the scoreboard specifically.

let myScoreboard = new Scoreboard()  // In file manage that manages the overall scoreboard (a different file than the one for damage).

init
    myScoreboard.addValue(damage)  // In file managing damage for the scoreboard.

    onDamageEvent() ->  // This goes in the file managing damage for the scoreboard.
        damageValues.incrementForPlayer(damagingUnit.getOwner(), damageAmount)

    onGameEnd() ->  // In file managing the overall scoreboard.
        myScoreBoard.show()
```

This lets you add columns without having to continually update the class definition or methods. Having two files avoids having one giant file tracking 20 different things, but does introduce some concerns, like how do you manage column order. Column order can be managed either by having one file add all the columns to the scoreboard in a hard-coded order, or by having the import order guarantee it, or by having the add column functionality include an index. There are trade-offs for each, so any of those work. You could also have a fixed order based on the column names or categories, but writing it out is probably best.

Jaccouille:
So `let damageValues = new ScoreboardValue(constructor args like picture for the column)` should init the damage column for all players and `damageValues.incrementForPlayer(damagingUnit.getOwner(), damageAmount)` update the scoreBoard entry?

I may have not precised this before, but I intended to keep playerScore class as it is for future purpose, such as registering the score to upload it, like it's done on [wc3stats.com](wc3stats.com) with replay parsing.

So far, I understood this
```
init
    let nameEntries    = new ScoreBoardColumnEntries()
    let levelEntries   = new ScoreBoardColumnEntries()
    let ElkEntries     = new ScoreBoardColumnEntries()
    let HawkEntries    = new ScoreBoardColumnEntries()
    let SnakeEntries   = new ScoreBoardColumnEntries()
    let WolfEntries    = new ScoreBoardColumnEntries()
    let BearEntries    = new ScoreBoardColumnEntries()
    let PantherEntries = new ScoreBoardColumnEntries()
    let DamageEntries  = new ScoreBoardColumnEntries()
    let HealingEntries = new ScoreBoardColumnEntries()
    let GoldEntries    = new ScoreBoardColumnEntries()
    let MeatEntries    = new ScoreBoardColumnEntries()
```

jonathanmclaus:
What is the point of having both ScoreBoard.wurst and PlayerScore.wurst?

Jaccouille:
> 
> 
> What is the point of having both ScoreBoard.wurst and PlayerScore.wurst?

Clarity, PlayerScore register scoring and ScoreBoard only handle the board part.

PR #597, opened 2020-09-07
Added a message displayed on mammoth & disco duck death

Jaccouille:
$changelog: Mammoth & Disco Duck death now display a message to all player.

People reported that Bosses death were too important to be left unknown for the the opponent team, they requested this change.

### DIFF
--- a/wurst/systems/entities/bosses/DiscoDuck.wurst
+++ b/wurst/systems/entities/bosses/DiscoDuck.wurst
@@ -10,6 +10,8 @@ import LocalObjectIDs
 import ItemPoolExtensions
 import LocalSoundUtils
 import StringExtensions
+import ColorUtils
+import PingMinimap
 
 public class DiscoDuck extends Hostile
     private static let unitTypeId = UNIT_DISCO_DUCK
@@ -56,6 +58,10 @@ public class DiscoDuck extends Hostile
         let pos = getPos()
         let pinion = pinions.placeRandomItem(pos)
         pinions.removeItemType(pinion.getTypeId())
+        // TODO : Wrap this up in a Boss entity class?
+        PlaySound(Sounds.warning)
+        printTimed(("The " + getUnit().getName() + " has been killed!").color("red".toColor()), 5)
+        pingMinimap(getUnit().getPos(), 5, COLOR_YELLOW, PingMinimapStyle.ATTACK)
         super.onDeath()
 
     override function postDeath()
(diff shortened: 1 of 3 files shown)


wurst/systems/entities/bosses/DiscoDuck.wurst
```diff
@@ -56,6 +58,10 @@ public class DiscoDuck extends Hostile
         let pos = getPos()
         let pinion = pinions.placeRandomItem(pos)
         pinions.removeItemType(pinion.getTypeId())
+        // TODO : Wrap this up in a Boss entity class?
+        PlaySound(Sounds.warning)
+        printTimed(("The " + getUnit().getName() + " has been killed!").color("red".toColor()), 5)
```
jonathanmclaus, line 63: This should use `format`.

I don't think it needs to go into `Boss` if you don't want to alert for every boss, but you shouldn't repeat the same logic in two different files. You could just keep it simple and add an event for unit death directly and check it against a custom set of IDs.

wurst/systems/entities/modules/Hostile.wurst
```diff
@@ -2,17 +2,22 @@ package Hostile
...
 import Sniff
 import UnitEntity
+import PingMinimap
+import ColorUtils
+import StringExtensions
 
 import initlater GreenFish
 import initlater Fish
 import initlater Hawk
+import initlater Mammoth
```
jonathanmclaus, line 19: Why is `initlater` required?

jonathanmclaus (changes requested):

wurst/systems/entities/modules/Hostile.wurst
```diff
@@ -62,4 +67,10 @@ public abstract class Hostile extends UnitEntity
         else
             udg_ANIMAL_CURRENT -= 1
 
+        if this instanceof Mammoth
+            or this instanceof DiscoDuck
```
jonathanmclaus, line 71: Can you change this condition to all be on the same line? I find this confusing at a glance because it looks like `or this instanceof DiscoDuck` is part of the block of the `if` structure, rather than the condition.

jonathanmclaus (changes requested):
OK. Not really ideal but the real problem is that using `UnitEntity` leads to diamond problems. Anyway,

jonathanmclaus:
I'll approve this but I think we should move off of the `UnitEntity` framework, at least for stuff like this.

jonathanmclaus (approved):

PR #598, opened 2020-09-09
Fixed bugs and made minor improvements.

Jaccouille:
$changelog: Fixed bug where Bee Hives would block vision.
$changelog: Fixed bug where basic materials would increase movement speed.
$changelog: Fixed bug where Cloak of Healing would affect unintended units.
$changelog: Fixed bug where Antimagic Shell would affect unintended units.
$changelog: Fixed bug where Dream Eater would fail to grant experience for kills.
$changelog: Fixed bug where Omnitower would be blocked from picking up any item.
$changelog: Fixed bug where tournament mode would disable tribe gold and buildings.
$changelog: Fixed bug where Frost Blast would have its cooldown reset immediately.
$changelog: Fixed bug where Thistles could be sold to merchants.
$changelog: Increased Dream Eater damage from 20 to 40 magic damage.
$changelog: Changed damage for Dream Eater to be reduced by spell resistance.
$changelog: Increased Shield Block minimum damage taken from 0 to 1.
$changelog: Reduced Hypnosis damage from 20 to 0.
$changelog: Corrected Bee Hive cost from 45 to 12.
$changelog: Corrected Workshop cost from 20 to 8 gold.
$changelog: Corrected Spirit of Darkness price from 25 to 12 gold.
$changelog: Updated Steel Coat tooltip to indicate that it can cast Cure All on the wearer.
$changelog: Updated Heat ability button position to match the grid position indicated by its hotkey.
$changelog: Changed gates to Mammoth pit to be destroyed instead of opened after grace period.
$changelog: Corrected charges per Antimagic and Drunk Potions from 1 to 2.
$changelog: Changed Stupefying Field to affect animals but not bosses.

Following changelog were forgotten in previous PR:
$changelog: Updated Battle Armor to grant 30% spell resistance.

Those 2 rocks felt lonely to me.


jonathanmclaus:
What prompts some of these number changes, like the Hypnosis damage removal?

Jaccouille:
> 
> 
> What prompts some of these number changes, like the Hypnosis damage removal?

Hypnosis : I wanted to centralize damage in 1 file, less code/tooltip to read/write, hypnosis sleep is already strong in itself I believe.
Shield Block minimum damage : An iron shield could block all damage from 10 omnitower, which felt... not nice to see, I literally had a teamfight in a base that lasted 3min (sage need a nerf) and it felt completely absurd to see all the traps & tower deal 0 damage.

The other number change were mistakes I made while rewriting item object definition.

jonathanmclaus:
What do you mean by all damage in 1 file?

Another note is that a lot of this stuff doesn't need to go into changelog. If we put everything in there, it might overflow when outputted in-game and is just noise otherwise. If there's no mechanical change, I wouldn't put it in. I'll edit it once I approve, just saying this for future note when the changelog gets too big.

Jaccouille:
Hypnosis & Dream eater are designed to work together, I actually didn't know how much EXACT damage those spells did, as Dream Eater use `unit.subhp(value)` to deal damage, it didn't take into account target magic resistance, which is 25%, therefore, those 2 spells combined deal 35 damage, I found this confusing so I decided to put the damage dealt in Dream Eater as pure damage, so now, you get a clearer picture of which spell sleeps the target and which spell deals the damage.

I actually cannot tell if the `subhp(value)` part is just plain crappy design from old developer or intended in order to deal pure damage.

wurst/objects/abilities/mage/hypnotist/DreamEater.wurst
```diff
@@ -10,18 +10,17 @@ import SimError
...
 let MANA_EATEN = 50.
-let DAMAGE = 20.
+let DAMAGE = 40.
 let MANACOST = 10
 let COOLDOWN = 20.
 
 let ANIMATION_NAME = "channel"
 let TOOLTIP_NORM = "Dream Eater"
-let TOOLTIP_EXTENDED = ("Steals an enemy dream restoring {0} Health Point and {1} Mana. Also reduces enemy's Mana by {2} and deals {3} damage. "+
+let TOOLTIP_EXTENDED = ("Steals an enemy dream restoring {0} Health Point and {1} Mana. Also reduces enemy's Mana by {2} and deals {3} pure damage. "+
```
jonathanmclaus, line 23: This is a big change so I'm not going to be picky about individual things right now when they can be bulk updated later, but in the future, let's try to be consistent about wording about points. ITT usually uses "energy" instead of "mana" and because it doesn't say "energy points" ever, I wouldn't say "health points" either. In this case, I would say "restoring X health and Y energy".

Regarding the "pure damage", I think that's just because it was coded loosely. It should probably use actual damage instead of setting HP so that AMS comes into play.

wurst/objects/abilities/mage/hypnotist/StupefyField.wurst
```diff
@@ -13,7 +13,15 @@ import BuffObjEditing
...
 
-let TARGET_ALLOWED = "enemies,friend,ground,hero,organic,self,vulnerable"
+let TARGET_ALLOWED = commaList(
+    TargetsAllowed.enemies,
+    TargetsAllowed.friend,
+    TargetsAllowed.ground,
+    TargetsAllowed.organic,
+    TargetsAllowed.self,
+    TargetsAllowed.vulnerable,
+    TargetsAllowed.nonancient
```
jonathanmclaus, line 23: What's the point of adding non-ancient?

Jaccouille: Should this spell affect mammoth or not? I'd say no, just to avoid further abuses.

wurst/objects/abilities/mage/hypnotist/DreamEater.wurst
Jaccouille: Writing down a proper norm may seem overkill, but it would help. I am dubious about the "energy" wording, new player don't immediately understand that energy == mana.

wurst/objects/abilities/mage/hypnotist/StupefyField.wurst
jonathanmclaus: OK, that's fine. I just wasn't sure that is what the effect was.

PR #599, opened 2020-09-23
Replaced acorn-based items with analogues appropriate to the setting.

Jaccouille:
$changelog: Replaced the cosmetics aspects of Acorn with Banana and Magic Acorn with Honeycomb.

Acorn was originally made because there existed an icon in the game for it, back in the old day, map size was limited and people didn't have the same download rate as today, I believe we can afford to add ~40kb of custom assets to make it more "immersive".


jonathanmclaus (changes requested):
Can you change `honeyComb` to `honeycomb`, including filenames? The capitalization is incorrect.

Did you have any thoughts for the smoke bomb recipe? I think this is fine for moving forward, but kind of silly now. I doubt it matters much considering no one makes smoke bombs.

PR #606, opened 2020-10-07
Fixed some pathfinding issues on NE and central islands.

### LINKED ISSUE #605: pathfinding makes you go around chokes sometimes
I checked a bit and I'm not sure why it's happening. This happens at very specific spots where the pathfinding just can't figure out how to get somewhere. Maybe some wc3 edge cases?

I tried fixing it by changing collision size and it works, but it's suboptimal since it changes how chokes are used (you can put 2 players in the chokes if collision size < 32). (e.g. do ..setCollisionSize(31) in createBasicTroll. Seems to be caused by tree and water herbs pathing. Any idea what's going on? Has that happened before/how was it fixed?

Some examples:
https://streamable.com/8guph1
https://streamable.com/0kjgb6
https://streamable.com/ffgttz

totony-misc:
$changelog: Widened some narrow passageways and chokepoints to reduce pathfinding miscalculations.

River + base close to seeds stash + middle top right ramp (#605)

Partial fix for #605

see
https://streamable.com/8guph1
https://streamable.com/0kjgb6
https://streamable.com/ffgttz


jonathanmclaus:
Please include pictures of before and after for each location being modified.

Also, streamable works but GIFs are better because GitHub supports hosting and displaying them directly.

totony-misc:
after-2
before-1
before-2
after-1

totony-misc:
middle ramp (on top of mastercrafter)

before:

before-3

after:

after-3

totony-misc:
shadow is fixed now

PR #607, opened 2020-10-07
Fixed miscellaneous bugs.

Jaccouille:
$changelog: Change Devour hotkey from F to T to avoid conflict with Sleep Outside.
$changelog: Fixed bug where where the quick-make for Bone Coat would fail incorrectly.

There was a weird bug with quickmake bone coat :
```
public class QuickMake
    let itemLists = new OwnedLinkedList<LinkedList<int>>()
    let specialHerb = new HashMap<int, int>
```

I don't really know how hashMap work under the hood, but for some reason the "specialHerb" hasMap would contain `ITEM_ELK_SKIN_COAT` as a key linked to a huge integer at some point, I moved the hashMap definition in the on cast quick make function scope so it get destroyed after the quickmake cast and avoid being "corrupted".

I also noticed that some quickmake had the `warstomp `orderId, which actually doesn't exist (anymore?). It didn't affect gameplay this much because wc3 would use `channel `orderId instead


jonathanmclaus:
I'm not 100% sure how it works in Wurst, but it looks like you were using a single instance for all of the `specialHerbs` objects, as you assign it outside the constructor.`

jonathanmclaus:
Remove the changelog file. That should be generated.

jonathanmclaus (changes requested):
Need to remove changelog.

wurst/systems/crafting/QuickMake.wurst
```diff
@@ -38,10 +35,11 @@ public class QuickMake
         let items = new OwnedLinkedList<item>()
         let inventory = caster.getInventory()
 
-        specialHerb.put(ITEM_BLUE_HERB, 0)
-        specialHerb.put(ITEM_YELLOW_HERB, 0)
-        specialHerb.put(ITEM_PURPLE_HERB, 0)
-        specialHerb.put(ITEM_ORANGE_HERB, 0)
+        let specialHerb = new HashMap<int, int>()
```
jonathanmclaus, line 38: This needs to be destroyed at the end.

jonathanmclaus (changes requested):

PR #609, opened 2020-10-07
Fixed bug where Elementalist would stop when attacked.

totony-misc:
$changelog: Fixed bug where Elementalist would potentially stop when damaged.

When overcharge was canceled (either by moving or stopping the channel), it would still order it to stop whenever it was attacked.

This is because time.stop() sets count to 0 and time.isLast checks if it is 1 (ideally this should check if its <= 1)

I'm surprised this hasn't been fixed before this is one of the main reasons i don't play elem.


Jaccouille (approved):
Nice catch.

jonathanmclaus (changes requested):
I want to review this to make sure the resources are cleaned up right.

PR #614, opened 2020-10-12
Fixed bug where central gate could be destroyed prematurely.

jonathanmclaus:
I am omitting this from the changelog because this bug was never reported and therefore it was likely never noticed.


jonathanmclaus:
This is an accidental duplicate of #601. The squash commit makes no changes and was removed afterwards via rebasing.

PR #619, opened 2020-10-26
Freshen base.w3x and, global variables + refactor spawn mechanics

totony-misc:
I remade a new w3x from scratch to remove legacy stuff that was left. This means anything but the custom doodads, the custom units, their placements, the map info and the regions are gone. e.g. udg_* are gone

I refactored the spawning system to make it more readable and removed weird semantics to know when animals/fishes died.

EDIT: That branch was to check if the pathing could be fix with a new w3x initially. It looks like the pathing problems are inherent to the terrain and doodads placement. E.g. only these 2 are required for the bug to happen.

For the spawn system, I kept the overall same logic, and haven't touched the semantics of the weighted item pool and weighted animal pool except for making it in a separate class.


totony-misc:
Take a look at the base.w3x to make sure its the same. I'm not sure why but the fog seems darker on the old one + it's raining on the old (but the settings are the same w/r to weather and fog)

totony-misc:
Also I remove wurst/systems/spawns/SpawnRateAdjustment.wurst cause it wasnt used and used some udg_* vars

totony-misc:
Also not sure why but the item count starts at -2? Although I haven't changed any reference to lowerCount

totony-misc:
> Also not sure why but the item count starts at -2? Although I haven't changed any reference to lowerCount

oh it must be when the tutorial gatherer crafts fire

jonathanmclaus:
This was closed to split into #620 and #621

PR #620, opened 2020-10-26
Cleaned up base map file.

totony-misc:
$changelog: Fixed bug where High Grass would not be displayed for some players.

I remade a new w3x from scratch to remove legacy stuff that was left. This means anything but the custom doodads, the custom units, their placements, the map info and the regions are gone. e.g. udg_* are gone, except a few

EDIT: That branch was to check if the pathing could be fixed with a new w3x initially. It looks like the pathing problems are inherent to the terrain and doodads placement. E.g. only these 2 are required for the bug to happen.


totony-misc:
some notes:
- Take a look at the base.w3x to make sure its the same. I'm not sure why but the fog seems darker on the old one + it's raining on the old (but the settings are the same w/r to weather and fog)
- I removed wurst/systems/spawns/SpawnRateAdjustment.wurst cause it wasnt used and used some udg_* vars
- currentItems starts at -2 because the gatherer in the tutorial makes a fire

totony-misc:
TODO: Remove the high grass portrait

here's some commits you can use to merge only some changes:
- 8fe53a0c885b75a67a932fa9bde350a91356a7dc - "add temporary global variables "
this has all the udg_* changes removed from the code. You can safely remove newbase.w3x from that commit and change the gitignore
- cf3470d14fe451cb8ca0332cb3c68d5d7c66cfd0 - "replace old .w3x with new clean-started w3x"
this has all the udg_* changes + the new base
- HEAD of this PR
this has all the udg_* changes + the spawn refactor (which further removes udg_current_* which were excluded from previous commits)

jonathanmclaus:
Considering you remade the `.w3x` from scratch, how certain are you that it's complete? There are lot of moving parts there and I think ITT was pretty good at minimizing them, but I'm not certain. For example, the current base has some imports under the Asset Manager. Are those included in the new one? By custom doodads and units, I thought you just meant the Object Manager entries, so we would be missing models after this.

totony-misc:
> Considering you remade the `.w3x` from scratch, how certain are you that it's complete? There are lot of moving parts there and I think ITT was pretty good at minimizing them, but I'm not certain. For example, the current base has some imports under the Asset Manager. Are those included in the new one? By custom doodads and units, I thought you just meant the Object Manager entries, so we would be missing models after this.

I am not 100% but fairly confident it's complete (apart from fog and rain). I manually copied the assets from the old base.w3x, but most of them seemed invalid. I also removed some that were unused in the wurst code (AFAIK). The only issue would be if the object manager has some unit tied to these asserts but it's unlikely and they are indicators thus removed by the code so it would not matter anyway.

Before:
image

Note that only Models/* Splats/* Textures/*, TeamGlow01a.blp, LoadingScreen.mdx and snake.blp are valid (in the actual mpq)

After:
image

I removed Splats/LightningData.slk because it did not seemed used by the code. This is the only one I'm unsure about. We can probably put that in the imports/ if you think its used.

I only added the HighGrass mdx since it's a doodads and you want to see in when placing it in editor. LoadingScreen.mdx probably doesn't need to be there either. Everything else is added by wurst.

jonathanmclaus:
You said in the initial description that the "regions are gone". But all the regions are still intact, right?

totony-misc:
yeah idk why i said that the regions are there

PR #621, opened 2020-10-26
Spawn system refactor + remove udg_current

totony-misc:
I refactored the spawning system to make it more readable and removed weird semantics to know when animals/fishes died.

For the spawn system, I kept the overall same logic, and haven't touched the semantics of the weighted item pool and weighted animal pool except for making it in a separate class.

currentItems starts at -2 because the gatherer in the tutorial makes a fire

### DIFF
--- a/wurst/systems/spawns/ResourceSpawns.wurst
+++ b/wurst/systems/spawns/ResourceSpawns.wurst
@@ -1,37 +1,76 @@
 package ResourceSpawns
 
-// Standard library imports:
-import ClosureTimers
-import ErrorHandling
 import HashMap
-import Interpolation
 import LinkedList
-import TerrainUtils
-import ClosureEvents
-
-// Third-party imports:
-import Lodash
-
-// Local imports:
-import BushSpawns
-import LocalObjectIDs
-import Game
-import GameTimer
-import GameConfig
-import GeometryUtils
 import Pool
-import LocalAssets
-import StringExtensions
-import ColorUtils
+
+import Composer
+import Lodash
+import LocalObjectIDs
+import ResourceSpawnsUtils
+import BushSpawns
+import GameConfig
+
+import ClosureEvents
+import Game
+import OnUnitEnterLeave
 import Toolkit
 
-let itemSpawnInfoMap = new IterableMap<int, SpawnInfo>()
-let animalSpawnInfoMap = new IterableMap<int, SpawnInfo>()
-let animalSpawnFunctions = new HashMap<int, VoidBiFunction<real, real>>()
-let islandSpawnList = new LinkedList<IslandSpawner>()
-let FISH_PER_AREA = 1
-var gameStartTime = 0.
+import ColorUtils
+import ClosureTimers
+import LocalAssets
+import StringExtensions
+import TerrainUtils
 
+let islandInfos = initIslandInfo()
+
+let itemSpawnInfoMap = new IterableMap<int, WeightScaler>()
+    ..put(ITEM_TINDER, new WeightScaler(450, 90, 860))
+    ..put(ITEM_CLAY_BALL, new WeightScaler(100, 185, 3600))
+    ..put(ITEM_STICK, new WeightScaler(300, 450, 360))
+    ..put(ITEM_FLINT, new WeightScaler(300, 250, 300))
+    ..put(ITEM_MANA_CRYSTAL, new WeightScaler(0, 160, 420))
+    ..put(ITEM_MUSHROOM, new WeightScaler(0, 120, 360))
+    ..put(ITEM_MAGIC, new WeightScaler(25, 25, 1))
+
+let animalSpawnInfoMap = new IterableMap<int, WeightScaler>()
+    ..put(UNIT_ELK, new WeightScaler(1000, 1000, 0))
+    ..put(UNIT_SNAKE, new WeightScaler(0, 45, 240))
+    ..put(UNIT_JUNGLE_WOLF, new WeightScaler(0, 90, 240))
+    ..put(UNIT_JUNGLE_BEAR, new WeightScaler(0, 45, 240))
+    ..put(UNIT_PANTHER, new WeightScaler(0, 45, 240))
+
+let fishSpawnInfoMap = new Pool<int>()
+    ..add(4, UNIT_HAWK)
+    ..add(2, UNIT_GREEN_FISH)
+    ..add(4, UNIT_FISH)
+
+// bookkeeper keys of spawninfos => current number of items spawned
+var currentFishes = 0
+var currentAnimals = 0
+var currentItems = 0
+
+let FISH_PER_AREA = 1
+
+let fishSpawnRects = new LinkedList<rect>()
+    ..add(gg_rct_out_1_1)
+    ..add(gg_rct_out_3_2)
+    ..add(gg_rct_out_1_2)
+    ..add(gg_rct_out_5_2)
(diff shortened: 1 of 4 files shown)


totony-misc:
This will conflict with #620 
Just use this branch's version and revert commit fad914faf0224b0437e56b7df8b7122d50930f2f

wurst/utils/Composer.wurst
```diff
@@ -0,0 +1,54 @@
...
+public class StaggeredComposer implements Composer
+    protected Composer Composer
+    protected int amount
+
+    construct(int amount, Composer Composer)
+        this.Composer = Composer
+        this.amount = amount
+
+    override function execute()
+        doPeriodicallyCounted(ANIMATION_PERIOD * 2, this.amount) fishHawkLoop ->
```
jonathanmclaus, line 31: This variable name looks like you missed it when you factored it out. Default is typically `cb`.

jonathanmclaus, line 31: Put the period under a `@configurable` global.

jonathanmclaus (changes requested):

wurst/systems/spawns/ResourceSpawnsUtils.wurst
```diff
@@ -0,0 +1,197 @@
...
+
+// Local imports:
+import Game
+import GameTimer
+import GameConfig
+import GeometryUtils
+import Pool
+
+function getRandomAllowedPoint(Pool<rect> allowedRegions, region restrainedSpawnRegion) returns vec2
+    let spawnRect = allowedRegions.random()
```
jonathanmclaus, line 18: This should have a global that controls the maximum number of points picked and an error message if it fails that many times. This should be a large number (e.g. 1000) that safeguards against an infinite loop when the region contains no valid points, which could happen during terrain changes and be very annoying to debug without this message.

wurst/systems/spawns/ResourceSpawnsUtils.wurst
```diff
@@ -0,0 +1,197 @@
...
+        this.allowedRegion = allowedRegion
+    
+    override function execute()
+        let spawnPoint = this.allowedRegion.randomPoint()
+        let targetID = this.units.random()
+    
+        // Spawn a unit of the selected type.
+        createUnit(
+            players[PLAYER_NEUTRAL_AGGRESSIVE],
+            targetID,
```
jonathanmclaus, line 71: There's really no point in creating local variables you only reference once, so I'd just move the above calls to replace `targetID` and `spawnPoint`.

wurst/systems/spawns/ResourceSpawnsUtils.wurst
```diff
@@ -0,0 +1,197 @@
...
+
+    construct(Pool<int> units, Pool<rect> allowedRegions, region disallowedRegions)
+        this.allowedRegions = allowedRegions
+        this.disallowedRegions = disallowedRegions
+        this.units = units
+    
+    override function execute()
+        let spawnPoint = getRandomAllowedPoint(this.allowedRegions, this.disallowedRegions)
+        let animal = this.units.random()
+        createUnit(players[PLAYER_NEUTRAL_AGGRESSIVE], animal, spawnPoint, randomAngle())
```
jonathanmclaus, line 89: Same as above - don't bother assigning locals just to use them once.

wurst/systems/spawns/ResourceSpawnsUtils.wurst
```diff
@@ -0,0 +1,197 @@
...
+    construct(real initialSpawnWeight, real finalSpawnWeight, real weightChangeTime)
+        this.initialSpawnWeight = initialSpawnWeight
+        this.finalSpawnWeight = finalSpawnWeight
+        this.weightChangeTime = weightChangeTime
+
+    function getTimeAdjustedSpawnWeight() returns real
+        if this.weightChangeTime == 0
+            return finalSpawnWeight
+        //Clamp ratio between 0 - 1 with min & max
+        let ratio = min(1, max(0, (getElapsedGameTime() - GAME_TIMER.getElapsed()) / this.weightChangeTime))
```
jonathanmclaus, line 117: I don't understand this line - why is `getElapsedGameTime` considered at all? That would be affected by mode and class selection times.

wurst/systems/spawns/ResourceSpawnsUtils.wurst
```diff
@@ -0,0 +1,197 @@
...
+            players[PLAYER_NEUTRAL_AGGRESSIVE],
+            targetID,
+            spawnPoint,
+            randomAngle()
+        )
+
+public class AnimalSpawnComposer implements Composer
+    protected Pool<int> units
+    protected Pool<rect> allowedRegions
+    protected region disallowedRegions
```
jonathanmclaus, line 79: This name is confusing. Using plural makes me think it's a collection of regions, but it's just one.

jonathanmclaus (changes requested):

wurst/systems/spawns/ResourceSpawnsUtils.wurst
totony-misc: My bad, I used the GAME_TIMER where it wasn't used before, but forgot to remove the -

wurst/systems/spawns/ResourceSpawnsUtils.wurst
```diff
@@ -0,0 +1,201 @@
...
+import GameTimer
+import GameConfig
+import GeometryUtils
+import Pool
+
+function getRandomAllowedPoint(Pool<rect> allowedRegions, region restrainedSpawnRegion) returns vec2
+    let spawnRect = allowedRegions.random()
+
+    var max = 1000 // Prevents infinite loop
+    while true and max >= 0
```
jonathanmclaus, line 22: Can't you use a for loop for this? And I meant a global to control the maximum number. Obviously you'll also need a local.

jonathanmclaus (changes requested):

wurst/systems/spawns/ResourceSpawnsUtils.wurst
totony-misc: what do you mean? while is a loop

ya i'm just not sure where to put that global, don't think it matters tbh since that should not happen and if it happens the spawn will just be broken for that one version

jonathanmclaus:
Doesn't the recipe island loop? how does that affect the item cap?

totony-misc:
> 
> 
> Doesn't the recipe island loop? how does that affect the item cap?

this item cap is checked before each item spawn:
```
                new ConditionalComposer(
                    condition,
                    new ItemSpawnComposer(itemPool, island.spawnRegionsPool, restrainedSpawnRegion)
                )
```

wurst/systems/spawns/ResourceSpawns.wurst
```diff
@@ -1,449 +1,305 @@
...
+    EmptyFunction<bool> condition = () -> (currentAnimals < gameConfig.getMaxAnimals())
+
+    // Make one spawner per island and combine at the end
+    let islandSpawners = new LinkedList<Composer>()
+    for island in islandInfos
+        let weightedPool = new WeightedPool(animalSpawnInfoMap)
+        weightedPools.add(weightedPool)
+        islandSpawners.add(
+            new StaggeredComposer(
+                (island.animalsSpawnCount * gameConfig.getFoodBase()).ceil(),
```
totony-misc, line 200: this won't take into account adjustSpawnRates - TODO

PR #632, opened 2020-11-11
Reduced health for Tent.

Jaccouille:
$changelog: Reduced health for Tent from 160 to 100.

Tent is pretty cheap, I believe its hp should be adjusted in relation to its cost.

### DIFF
--- a/wurst/objects/units/units.wurst
+++ b/wurst/objects/units/units.wurst
@@ -1979,9 +1979,9 @@ import LocalAssets
 
 @compiletime function create_w3u_n004()
 	let _def = createObjectDefinition("w3u", 'n004', 'ntn2')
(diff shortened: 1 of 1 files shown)


jonathanmclaus:
I was thinking something more like 40 - you can't kill it without wasting time in a rush but not feasible after the enemy has axes. Do you think that's too low?

jonathanmclaus:
I don't think we should let it be pack. Let's try reducing the HP to 100 for now and see if it's still abusable.

jonathanmclaus (changes requested):
Revert the change for abilities.

Jaccouille:
And change its armor type from fortified to large? Fortified reduces hero damage by 50%, if you have 20 damage, it would take you 10 hits to kill a tent.

PR #633, opened 2020-11-11
Reduced snake poison effects.

Jaccouille:
$changelog: Reduced movement speed reduction for Snake poison from 35% to 20%.
$changelog: Reduced duration for non-heroes for Snake poison from 300 to 150 seconds.

Snake poison is too annoying, 2min is way too long, I don't understand the purpose of this duration, I reduced it to 30 seconds so if you get attacked by a snake when you are escaping/fighting with other troll, it still feels like it has an impact.

### DIFF
--- a/wurst/objects/abilities/BaseMapAbilities.wurst
+++ b/wurst/objects/abilities/BaseMapAbilities.wurst
@@ -109,8 +109,8 @@ import ObjectIds
         ..setDamagePerSecond(1, 0.0)
         ..setMovementSpeedFactor(1, 0.35)
         ..setAttackSpeedFactor(1, 0.07)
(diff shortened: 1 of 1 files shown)


Mephistophelesss:
I'm not against the idea, but maybe changing the slow percentage instead of duration? I feel like fighting with snake was meant to be aids, sure you get exp out of it but you really don't want to get bit. If it last 30 seconds, you sleep after killing it thats 20 seconds left, snake wont feel like much of a threat anymore.

jonathanmclaus:
I agree with @Mephistophelesss. This change would make snake basically the best source of free experience. Reducing the movement speed reduction from 35% to 20% or even 10% would be better I think.

PR #636, opened 2020-11-14
Added a command to enable legacy hotkeys.

Jaccouille:
$changelog: Added the -lh command to enable legacy hotkeys for some troll abilities, which must be used prior to selecting a troll.

Some old player do not wish to play with the new spell hotkeys which were implemented after 3.2, the idea was to have a consistent grid-like hotkeys mapping for abilities, instead of having hotkeys being the first letter of the spell, like H for Hypnosis.

In order to accomodate those player, I added a -lh command which has to be entered 10 seconds after the game start (after troll selection), it will swap current mage spell with a duplicate one who just has a different hotkey.

I added a condition in the autoSkill logic to update the legacy spellbook instead of current one in case -lh has been enabled. I stocked a "legacyHotkey" boolean in a config class who has an instance mapped for each player, guess it could be used for something else later on.

I didn't think twice about implementation as I am not really in the mood to design a "proper" configuration system, I just want to make a hotfix so those player stop complaining about hotkeys.


wurst/systems/core/PlayerConfig.wurst
```diff
@@ -0,0 +1,35 @@
+package PlayerConfig
+
+import HashMap
+import GameMode
+
+public class PlayerConfig
+    static let playerToConfigMap = new HashMap<player, PlayerConfig>()
```
jonathanmclaus, line 7: This should use `thistype` in the generic.

jonathanmclaus (changes requested):

wurst/systems/core/PlayerConfig.wurst
Jaccouille: I don't understand, do you mean this? This doesn't work cannot infer type for key/value.
```
public class PlayerConfig
    static let playerToConfigMap = new HashMap<player, this>()
```

Jaccouille: Nevermind, it should be as you wrote `thistype `not `this`, my bad.

jonathanmclaus:
I pushed work in progress to have a backup while debugging. It's not worth looking at in its current state.

PR #638, opened 2020-11-14
Reverted health and armor type for some buildings.

Jaccouille:
$changelog: Reverted health for Mixing Pot, Workshop, and Witch Doctor's Hut from 250 to 260, 275, and 325, respectively .
$changelog: Reverted armor type for Mixing Pot and Workshop from Fortified to Heavy.

I've noticed those changes compared to 3.2 version, I believe this is an oversight from rewriting the troll buildings definition. Building HP should scale with their price.

### DIFF
--- a/wurst/objects/units/TrollBuildings.wurst
+++ b/wurst/objects/units/TrollBuildings.wurst
@@ -63,6 +63,9 @@ function createBuilding(int newId, int oldId) returns BuildingDefinition
         ..setTooltipBasic("You can mix herbs inside this pot.")
         ..setTooltipExtended("This pot is so big it is able to hold large amounts of boiling water. Inside, herbs can be mixed into potions.")
         ..setName("Mixing Pot")
+        ..setHitPointsMaximumBase(260)
+        ..setArmorType(ArmorType.Large)
+        ..setDeathTimeseconds(1)
 
 @compiletime function createTannery() returns BuildingDefinition
     return createBuilding(UNIT_TANNERY)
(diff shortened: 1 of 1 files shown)


jonathanmclaus:
It wasn't oversight. I just didn't think the difference was impactful enough for people to even play around. What do Large and Heavy armor types do? Can we use a weaker armor type for fire and tent?

Jaccouille:
> 
> 
> It wasn't oversight. I just didn't think the difference was impactful enough for people to even play around. What do Large and Heavy armor types do? Can we use a weaker armor type for fire and tent?

Fortified currently reduce Hero damage by 50%, normal by 30%, pierce by 75%, Large armor takes 100% damage from those 3 attack types. I noticed it but I just didn't care this much, I am making the change, here again, to answer the current feedback.

totony-misc:
image
http://classic.battle.net/war3/basics/armorandweapontypes.shtml

totony-misc:
imo this is not an issue though

wurst/objects/units/TrollBuildings.wurst
```diff
@@ -63,6 +63,9 @@ function createBuilding(int newId, int oldId) returns BuildingDefinition
         ..setTooltipBasic("You can mix herbs inside this pot.")
         ..setTooltipExtended("This pot is so big it is able to hold large amounts of boiling water. Inside, herbs can be mixed into potions.")
         ..setName("Mixing Pot")
+        ..setHitPointsMaximumBase(260)
+        ..setArmorType(ArmorType.Large)
+        ..setDeathTimeseconds(1)
```
jonathanmclaus, line 68: What happens without this?

Jaccouille: Pot death animation plays twice because the model animation last 1 seconds. Have to sync it with the actual unit definition.

PR #639, opened 2020-11-15
Fixed scoreboard timer for observers.

Jaccouille:
$changelog: Fixed scoreboard timer for observers.

When I rewrote scoreboard, I made it so it only update on event such as kill/item sold, but I forgot to periodically update the board timer.

### DIFF
--- a/wurst/systems/boards/ScoreBoard.wurst
+++ b/wurst/systems/boards/ScoreBoard.wurst
@@ -238,6 +238,9 @@ init
                     for member in tribe.getMembers()
                         scoreBoardEntries.get("Level")
                         .updateEntryIcon(member.getName(), member.getTrollIcon())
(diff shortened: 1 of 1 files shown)


wurst/systems/boards/ScoreBoard.wurst
```diff
@@ -238,6 +238,9 @@ init
                     for member in tribe.getMembers()
                         scoreBoardEntries.get("Level")
                         .updateEntryIcon(member.getName(), member.getTrollIcon())
+                doPeriodically(1) (CallbackPeriodic cb) ->
+                    let title = getGameTimersBoardTitle()
```
jonathanmclaus, line 242: There's no point in having this local. `socreBoard.board.setTitle(getGameTimersBoardTitle())` is succinct enough I think

jonathanmclaus (approved):

PR #641, opened 2020-11-17
Increased mana drain per hit on Mage Masher.

theQuazz:
$changelog: Increased mana drain per hit on Mage Masher from 6 to 10.

With mage and priest having a mana pool of 350 mage masher feels like a butter knife. This will give it slightly more power. Longer term we may want to consider % drain.


Jaccouille:
As discussed in Discord, it should be fairly easy to implement a custom mana burn which would apply a % of max mana instead of flat amount, but I can agree with this as a quick fix.

PR #642, opened 2020-11-17
Magic Tweaks

theQuazz:
- replace magic's 30% life/mana random option with packing all fires on the map
- reduce magic exp gain from 500 to 300

Quantum says he will stop playing if we do the first change. It's really just a meme. I do think that the exp change is a good idea though.

### DIFF
--- a/wurst.build
+++ b/wurst.build
@@ -1,3 +1,4 @@
+---
 projectName: Island Troll Tribes
 dependencies:
 - https://github.com/wurstscript/wurstStdlib2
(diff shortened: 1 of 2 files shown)


Jaccouille:
You may want to update the effect created upon the caster, it is currently death coil. I suggest adding an effect upon fire being packed by magic effect so player actually understand why the fire has been packed, unless you want to really want to mess with them.

theQuazz:
@Jaccouille 500 exp is a lot. I don't think that I'll actually try to merge the fire change, too memey.

jonathanmclaus:
Yeah, the fire effect seems really random. I'm not a fan of experience being on magic at all, especially if it's going through the experience system.

jonathanmclaus:
My only thoughts on magic rework would be to still have a variety of effects, but make them more deterministic, such as being determined by time of day.

jonathanmclaus:
This is both random and doesn't address the RNG factor.

PR #647, opened 2020-11-19
Fixed bug where scoreboard displayed prematurely in 3+ team games.

Jaccouille:
$changelog: Fixed bug where the scoreboard would be displayed prematurely in games with many tribes.

I forgot to specify player in the scoreboard display method upon a tribe defeat, therefore it displayed the scoreboard for every one when 1 tribe was defeated.
Also fixed the square icon next to the tribe name which now turns red when a tribe is defeated.

### DIFF
--- a/wurst/systems/core/Tribe.wurst
+++ b/wurst/systems/core/Tribe.wurst
@@ -17,6 +17,7 @@ import PlayerExtensions
 import StringExtensions
 import initlater TribeBoard
 import initlater Boards
+import initlater ScoreBoard
 
 let listeners = CreateTrigger()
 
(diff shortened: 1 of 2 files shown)


wurst/systems/core/Tribe.wurst
```diff
@@ -125,7 +126,8 @@ public class Tribe
             TribeBoard.findForPlayer(member).board.display(member, false)
 
             // Enabling the score board, maximizing it so the player doesn't miss it
-            scoreBoard.board..display(true)..maximize(member)
+            // scoreBoard.board..display(true)..maximize(member)
```
jonathanmclaus, line 129: Why leave this commented? Just remove if it the logic was a bug.

jonathanmclaus (changes requested):

wurst/systems/core/Tribe.wurst
Jaccouille: You're right, forgot about that.

PR #650, opened 2020-11-20
Reverted level requirements for supersubs.

theQuazz:
$changelog: Reverted level requirements for supersub to level 9.


Jaccouille:
Those changes were made because Juggernaut funneling was a thing nearly every game in 2V2 & 3v3 with other class being played as "the juggernaut support", it was reported that playing thief or priest just to be a support was quite boring and repetitive.

I believe super sub design is flawed, it was just a secret in the first place meant to be "off-balance" due to the RNG medallion drop. I don't get why you can access super sub without medallion in the first place, that's just pushing player to farm XP rather than looking for fight after subbing.

theQuazz:
@Jaccouille makes sense why the change was made, but I don't think that it's the right change. I totally agree with your assessment of supersub being flawed, and I too would like to make medallion required to get it, but that is a larger rework than I'm aiming for in this PR. I think @Mephistophelesss had good intentions in making super sub more accessible, and I think that it has lead to countless crazy fun games, but perhaps we should make it require medallion again. I plan on contributing more ideas towards that objective, but not on GitHub right now.

PR #651, opened 2020-11-20
GoK experience gain from 50% -> 25%

theQuazz:


jonathanmclaus:
Can we withdraw this? It's a pretty minor change so reimplementing it later isn't hard and I think we won't end up using this now that all level requirements were raised to 9 in #650

theQuazz:
Yep seems to be more balanced now

PR #656, opened 2020-11-24
Updated -obs command to end mode selection when appropriate.

Jaccouille:
Ending the game mode selection when the player from the first slot type -obs, this way people don't have to wait when the player picking mode goes observer.

### DIFF
--- a/wurst/systems/commands/Commands.wurst
+++ b/wurst/systems/commands/Commands.wurst
@@ -114,6 +114,8 @@ init
             triggerPlayer
         )
         triggerPlayer.makeObserver()
+        if triggerPlayer == players[0]
+            GameMode.endModeSelection(true)
 
     registerCommandAll("obs-new") (triggerPlayer, args) ->
         printTimedToPlayer(


jonathanmclaus:
I think it would make more sense to delay the `obs` action until after mode selection ends. What do you think? Basically I think the changes from modes should be finalized once mode selection finishes.

theQuazz:
That will make troll selection a bit messier because then you might need to undo someone picking a troll. Also the enemy team might think they're playing against a different number of characters so choose a certain composition and then someone could obs at the last second, throwing off their plan. It was very intentional for me to only allow `-obs` during mode selection, which should be thought of as game initialization.

theQuazz (approved):

jonathanmclaus (changes requested):
I think we should discuss how this should work.

jonathanmclaus:
@theQuazz I meant that `-obs` would only be allowed during mode selection, but that it would take effect when mode selection ended, preventing the host from accidentally changing to observer before finishing modes.

theQuazz:
@jonathanmclaus ah that makes sense

jonathanmclaus:
Right now it's slightly helpful to immediately become an observer because you see the update in the allies screen, but I thought at some point there was a plan to switch off the standard observer behavior for WC3, in which case that benefit doesn't apply.

theQuazz:
I tried but it wasn't successful. We can keep trying though if you want.

jonathanmclaus:
What was the problem? I don't remember if you tried before tribes were always a thing. I think we could just have a pool of players without a tribe, which have shared vision from all players, always a scoreboard, no troll, and are ignored for game ending conditions. The last 2 behaviors should happen naturally from not having a tribe.

theQuazz:
Try `-obs-new` in game

jonathanmclaus:
I think we should switch to new-style observers but this is a good workaround in the meanwhile.

PR #664, opened 2020-12-05
Fixed bug where Track failed to cast for Trapper.

totony-misc:
$changelog: fixed trapper's tracking trap spell so it doesn't error when activated

### DIFF
--- a/wurst/objects/abilities/scout/Trap.wurst
+++ b/wurst/objects/abilities/scout/Trap.wurst
@@ -8,7 +8,7 @@ import ClosureEvents
 import ClosureForGroups
 import ErrorHandling
 import HashMap
-import DummyCaster
+import InstantDummyCaster
 import Orders
 
 // Third-party imports:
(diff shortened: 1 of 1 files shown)


totony-misc:
dont merge ill delete when it works

wurst/objects/abilities/scout/Trap.wurst
```diff
@@ -19,6 +19,8 @@ import UnitExtensions
 import LocalObjectIDs
 import ToolTipsUtils
 
+import Track
```
jonathanmclaus, line 22: This sort of dependency is bad as I don't want to worry about circular imports. I would define the track in that file directly and just reference the ability ID. It's messy in a different way, but I think this should be the standard because WC3 already related independent object purely on their ID. It also makes it easier to see the implementation for an ability.

jonathanmclaus (changes requested):
Waiting on file re-org.

jonathanmclaus:
Closed in favor of #674

PR #668, opened 2020-12-05
Froze order IDs for abilities based on Channel.

Jaccouille:
I set the base order Id for all channel ability preset that were still relying on the automatic orderId attribution from the ChannelAbilityPreset class.

The issue was that the automatic setting could result in a conflict between spells and those would not cast properly.


jonathanmclaus:
I would add that the conflict only arises (right now) due to the banning of order IDs not working properly, as the generator can and does mismatch target types and order IDs if we ban all of the order IDs used in basic abilities for this map.

Even if that is fixed, this might still be a problem because there are enough abilities based on Channel that we need more than the total number of available order IDs, meaning the generator will loop back around and a unit may have abilities with the same order ID generated from multiple iterations on that loop.

PR #669, opened 2020-12-06
Change Beast Master subclass tame pet system to Locust Swarm variant (POC)

theQuazz:


wurst/objects/abilities/beastMaster/SummonBeast.wurst
```diff
@@ -0,0 +1,90 @@
...
+let TOOLTIP = "Summon a beast to fight along side you"
+let TOOLTIP_EXT = "Each level will increase the ferocity of the animal fighting along your side"
+
+let DAMAGE_BASE = 7
+let DAMAGE_RANGE = 3
+let DAMAGE_PER_LEVEL = 3
+let COOLDOWN = 60.
+let DURATION = 30.
+
+class ElkLocust extends AdultElkAttackerDefinition
```
jonathanmclaus, line 24: I don't see the point of using a class type per ability definition and have been trying to move away from that, like how @Jaccouille did in #668

jonathanmclaus (changes requested):

wurst/objects/abilities/beastMaster/SummonBeast.wurst
Jaccouille: We realized it wasn't good design as it lacked flexibility and might end up with "big" constructor prototype which were hard to read, [this](https://github.com/island-troll-tribes/island-troll-tribes/blob/cc1bc2981f5f7c11c3bdf91e9c73fee74b62f247/wurst/objects/units/TrollBuildings.wurst) is an example of what @jonathanmclaus came with, you end up with more line but it's easier to update if you need to alter an attribute for a specific object definition.

theQuazz: I don't mind switching it, seems like a reasonable conclusion. But before spending any more time on this idea we should determine which course of action we want to take on this rework.

jonathanmclaus: I think it works better than the current system, so still worth merging it even if it's temporary.

Jaccouille: If you want to talk about course of action, you might want to describe the context in the PR, because I don't understand :
1. What is the current issue
2. How does this solve the issue or what are the expected result, iirc it's an ability who spawns a locust unit for X seconds, which attack nearby enemies, I could label it as "uncontrolled damage source" and I kind of fail to see where the fun part in playing this.

jonathanmclaus: @Jaccouille , I think the core problems are (in my opinion):

1. Bear is a big power spike, as it is stronger and not really balanced.
2. Any pet makes killing hostiles trivial, as they have free healing and revive.
3. You can have the pet kill items on the enemy island with minimal risk. Even if the enemy finds it, they have to waste their time killing it and again, this costs the BM nothing.

I do think that is not as fun, but it solves all of these immediately and still is relevant to the class theme. I'd like to address these better in the long run though.

Jaccouille: Why not nerf bear stats and have pets give XP on death?

theQuazz: Like Jon said, right now it's not balanced, and generally I think it would be difficult to balance Shapeshifter having a pet with how powerful the beast forms should be. Thus, in order to have ability continuity from the base class to the subclass, I would like to switch to the locust abilities. Yes, this might make BM a little more right-clicky, but items give you so much interactivity that I don't think this is a problem. I plan on keeping pets in their current form on Druid, and eventually I'll work on improving the pet system for Druid/Rendo.

Jaccouille: Still believe this is less fun/interactive than #569 but you're free to test w/e you want as I trust you have bigger plan.

theQuazz: @Jaccouille your PR can still be merged into the plan for Druid, wouldn't you agree?

Jaccouille: I am not against this suggestion and would gladly spent some time to adjust/think more about it, however I am on a tight schedule at this moment.

wurst/objects/units/TrollUnitTextConstant.wurst
```diff
@@ -56,19 +56,17 @@ public constant BASIC_TROLL_SPELLS = commaList(ABILITY_PANIC,ABILITY_SLEEP_OUTSI
...
-// The first ability is just a placeholder that is stripped during creation of the class menu.
-// TODO: Fix this hack by properly listing out abilities and class selection separately.
-public constant HERO_SPELLS_BEAST_MASTER = commaList(ABILITY_SPIRIT_BEAST,     ABILITY_SHAPESHIFTER,   ABILITY_DRUID)
-public constant HERO_SPELLS_THIEF        = commaList(ABILITY_CLOAK,            ABILITY_ESCAPE_ARTIST,  ABILITY_CONTORTIONIST,ABILITY_ASSASSIN)
-public constant HERO_SPELLS_SCOUT        = commaList(ABILITY_REVEAL_ID,        ABILITY_OBSERVER,       ABILITY_TRAPPER,      ABILITY_SPY)
-public constant HERO_SPELLS_GATHERER     = commaList(ABILITY_ITEM_RADAR,       ABILITY_RADAR_GATHERER, ABILITY_HERB_MASTER,  ABILITY_OMNIGATHERER)
+public constant HERO_SPELLS_MAGE         = commaList(ABILITY_MAGE_SPELLBOOK,   ABILITY_ELEMENTALIST,   ABILITY_HYPNOTIST,     ABILITY_DEMENTIA_MASTER)
+public constant HERO_SPELLS_PRIEST       = commaList(ABILITY_PRIEST_SPELLBOOK, ABILITY_BOOSTER,        ABILITY_MASTER_HEALER, ABILITY_SAGE)
+public constant HERO_SPELLS_BEAST_MASTER = commaList(ABILITY_SUMMON_BEAST,     ABILITY_SHAPESHIFTER,   ABILITY_DRUID)
```
jonathanmclaus, line 61: Keep the comment. It is still applicable and should be noted.

wurst/objects/abilities/beastMaster/SummonBeast.wurst
```diff
@@ -0,0 +1,90 @@
...
+        ..setCheckDependencies(false)
+        ..setTooltipNormal(1, TOOLTIP)
+        ..setTooltipNormalExtended(1, TOOLTIP_EXT)
+        ..setHeroAbility(true)
+
+@compiletime function createSummonBeastSpellBase()
+    createSummonElkSpell(ABILITY_SUMMON_BEAST)
+    new ElkLocust(UNIT_LOCUST_BEAST_ELK)
+    new WolfLocust(UNIT_LOCUST_BEAST_WOLF)
+    new BearLocust(UNIT_LOCUST_BEAST_BEAR)
```
jonathanmclaus, line 90: Fix your VS Code settings and keep the line break.

wurst/objects/abilities/beastMaster/SummonBeast.wurst
```diff
@@ -0,0 +1,90 @@
...
+        super(unitId)
+        this
+            ..setName("Elk")
+            ..setNormalAbilities("Aloc")
+            ..setAttack1DamageSidesperDie(DAMAGE_RANGE)
+            ..setAttack1DamageBase(DAMAGE_BASE)
+            ..setSpeedBase(522)
+            ..setTurnRate(0.3)
+
+class WolfLocust extends AdultJungleWolfDefinition
```
jonathanmclaus, line 35: I think these sort of constructs should be dropped in favor of chain function calls relating to `AbilityDefinition` but this is OK for now.

jonathanmclaus (changes requested):

PR #680, opened 2020-12-16
Expanded support for cosmetics to include attachments; added hat for randoming.

totony-misc:
I've been asking for hats in this game for a bit lets bring TF2 vibes here.

This adds a per-player perk that gives hats to their trolls. Might wanna do something else than hats but I think it's pretty nice.
Perk utils might also be nice for skins.

Priest is blacklisted because its head attachment position is dumb (any idea how to fix it? I'll try a bit more later)


totony-misc:
example for how it looks like on mage
image

It stacks with custom skins for now to double one-up other players.

PR #705, opened 2020-12-26
Fixed Neutral Building ability.

Jaccouille:
Should have been merged with #585, I forgot about it.

### DIFF
--- a/wurst/objects/abilities/SelectAbilities.wurst
+++ b/wurst/objects/abilities/SelectAbilities.wurst
@@ -15,14 +15,16 @@ import LocalAssets
 
 // Used for fire so enemies can use pack up
 @compiletime function createNeutralBuildingSelectAlliedAndEnemy() returns AbilityDefinitionNeutralBuilding
-    return new AbilityDefinitionNeutralBuilding(ABILITY_SELECT_ALLIED_AND_ENEMY)
-        ..setArtCaster("")
+    return new AbilityDefinitionNeutralBuilding(LocalAbilityIds.selectHero)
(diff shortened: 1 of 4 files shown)


wurst/objects/abilities/SelectAbilities.wurst
```diff
@@ -15,14 +15,16 @@ import LocalAssets
 
 // Used for fire so enemies can use pack up
 @compiletime function createNeutralBuildingSelectAlliedAndEnemy() returns AbilityDefinitionNeutralBuilding
-    return new AbilityDefinitionNeutralBuilding(ABILITY_SELECT_ALLIED_AND_ENEMY)
-        ..setArtCaster("")
+    return new AbilityDefinitionNeutralBuilding(LocalAbilityIds.selectHero)
         ..setCasterAttachmentPoint("")
-        ..setActivationRadius(1, 150.0)
+        ..setActivationRadius(1, 200.0)
```
jonathanmclaus, line 20: Does the radius actually matter?

Jaccouille: When an unit enter the radius, it gains vision of the neutral unit, if you put 1000 as radius, you would be able to see what's around the neutral from 1000 range afar, I didn't confirm but I guess the neutral unit share vision with the player once a unit enter this radius. 200 was the original value in abilities.wurst

jonathanmclaus: I've never noticed gaining vision from a fire.

PR #706, opened 2020-12-26
Added back dropped change.

Jaccouille:
Should have been fixed in #585.


jonathanmclaus:
This change was in #633 and discussed there. Just dropped in the merge from #585.

PR #707, opened 2020-12-26
Fixed hotkey for sleep.

Jaccouille:
Another mistake from #585 I didn't see.


jonathanmclaus:
Leaving this out of changelog because it's only in winter version.

PR #713, opened 2020-12-28
Fixed cloak stats bonus ability & spiked trap ID

Jaccouille:
Another mistake from #585 

Question for @jonathanmclaus , I used the `AIx3 `ability which isn't registered in the AbilityIds class from the wurst stdlib, do you want it to be added to the LocalAssets  instead of adding a line to the LocalObjectIds?


jonathanmclaus:
If that is an ability built into the game, then the standard library should be updated and we should use it. I think we should try to avoid using temporary band-aids, such as in `LocalAssets` for assets that aren't local, unless we can't contribute to the standard library for some reason.

Jaccouille:
This PR is deprecated since spiked trap ID has been fixed, see #727 for cloaks ability Id fix.

PR #714, opened 2020-12-28
Organized folder for imported models.

Jaccouille:
I've been having trouble browsing through this import folder, decided to sort it a bit for clarity.


jonathanmclaus:
Is there any logic behind sorting the models other than what they're currently used for? For example, I treated the `Crown.mdx` model as an item, but it's only used on an ability. And you can use the same model for both an ability and unit if you wanted to.

Jaccouille:
> 
> 
> Is there any logic behind sorting the models other than what they're currently used for? For example, I treated the `Crown.mdx` model as an item, but it's only used on an ability. And you can use the same model for both an ability and unit if you wanted to.

Well WC3 follow this logic, models used as items are put in `Objects/InventoryItems/` for example, I followed it because I usually browse through wc3 assets to edit them, this is more familiar to me.

I'd put the `crown.mdx` & `strawhat.mdx` models in abilities because they're used as effect.

PR #715, opened 2020-12-28
Standardized text tags.

Jaccouille:
$changelog: Added a text tag to visualize the mechanical effects of Mix Energy and Mix Heat.

I find it hard to see the mix spell effect, added a texttag to improve clarity


jonathanmclaus:
Can you add one for the other two (Health and Heat)?

wurst/objects/abilities/priest/unsub/MixSpells.wurst
```diff
@@ -91,6 +106,8 @@ function mixMana(unit caster, unit target)
     let casterMana = caster.getMana()
     let newMana = (targetMana + casterMana) / 2.00
 
+    createFadingTextTag(target, getMixDiffStr((targetMana - newMana).toInt()), COLOR_TURQUOISE)
+    createFadingTextTag(caster, getMixDiffStr((casterMana - newMana).toInt()), COLOR_TURQUOISE)
```
jonathanmclaus, line 110: Why not use `COLOR_ENERGY` or what is used for Mana Burn?

Jaccouille: I used turquoise as it brighter, thought it would be easier to notice

jonathanmclaus:
Just want to confirm that this is intended to display text tags without considering the units' maximum mana and heat. If you mix a level 4 priest at 220 energy with a level 1 troll at 200 energy, the text tag will indicate that the target troll gained energy even though they were at cap. I think this makes sense, as stuff like critical strike displays the full damage that could have been done even if the target was at 1 HP before being hit.

jonathanmclaus:
Changed Panther Instinct to use "¬" instead of "Dodged". I think this is more inline with how Warcraft 3 stylizes text tags - they usually just use a symbol except for "miss" which doesn't display damage. I'd like to have used "~" but it doesn't format well in WC3.

PR #728, opened 2021-01-04
Fixed movement for Ancient Hydra.

Jaccouille:
$changelog: Fixed bug where the Ancient Hydra would occasionally return to its spawn.

Current hydra behavior : every 4 seconds, check if its acid breath ability is on cooldown, if not, check if enemies are around, if yes cast ability, if not, attack order on mammoth pit. If ability is on cooldown, do nothing. 

So when hydra cast its spell, which has 10 second cooldown, it is ordered to move back to its original position due to neutral AI, which order move every ~5seconds, that's why it goes back, then once he can cast its ability again, he gets ordered to move at mammoth pit.

Fix: The unit.setPos(vec2) actually reset the neutral unit guarding position, I called it everytime hydra gets ordered to move at mammoth pit, resetting its guarding position to its current position, next time neutral AI order it to move, it won't go straight back at its spawning position.


wurst/systems/entities/bosses/AncientHydra.wurst
```diff
@@ -59,6 +59,9 @@ public class AncientHydra extends Hostile
             end)
 
         if frostReady
```
jonathanmclaus, line 61: Does this only happen if `frostReady` is true and there isn't a nearby unit to use breath of frost (or the equivalent ability) on?

Jaccouille: Yes

PR #732, opened 2021-01-23
Mage masher nerf

totony-misc:
having played a lot on the new patch, 10 mana burn is too much

### DIFF
--- a/wurst/objects/items/Armory/AxesDefinition.wurst
+++ b/wurst/objects/items/Armory/AxesDefinition.wurst
@@ -52,8 +52,8 @@ let TT_COMMON = ""+
     "Increases the attack damage of the unit by " + "{0}".color(RED_COLOR) + " when worn. "+
     "Wielder can cut trees."
 
-let TT_MM = (TT_COMMON +
-    " Can cast silence which has a {0} area of effect, lasting {1} seconds "+
+let TT_MM = TT_COMMON +
+    (" Can cast silence which has a {0} area of effect, lasting {1} seconds "+
     "with a {2} second cooldown. ")
     .format(
         SILENCE_AOE.toToolTipLightBlue(),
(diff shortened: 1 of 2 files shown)


jonathanmclaus:
Can you make the tooltip for Mage Masher reference this value (and also fix the other incorrect parts)?

jonathanmclaus:
Also I don't think this really fixes the issue for team games, so changes introducing a scaling value for stacking (similar to spears) would be welcome.

totony-misc:
> Also I don't think this really fixes the issue for team games, so changes introducing a scaling value for stacking (similar to spears) would be welcome.

im not sure people actually want that but im not planning on working too much on itt right now

theQuazz:
Scaling it as I suggested would be better

jonathanmclaus:
Your way doesn't scale based on players either though.

theQuazz:
No, I just don't think that it's an issue that should be solved for. You should have to have great position management as a caster. Implementing a solution similar to spears more broadly would be a shame in my opinion because I see the spears change as a shim rather than a long term change.

jonathanmclaus:
Can't really call the spears change a shim when it's been there for 3
years. I think people are familiar enough with the system that it would
work for mana burn.

theQuazz:
I'd still call it a shim. My goal has been to introduce more items which increase your hp pool thus making spears ineffective. But progress on that has been delayed for various reasons, primarily lack of consensus. Hence why we've been stuck with this shim. That doesn't make it not a shim.

totony-misc:
> Scaling it as I suggested would be better

what did you suggest? % of mana burn?

theQuazz:
% of current mana with a minimum value

jonathanmclaus:
Conflicts with #750. Will open a new PR if needed.

PR #733, opened 2021-01-23
Add back overcharge and brambles

totony-misc:


jonathanmclaus:
I'm not just going to revert this change. Talk to wink about about the proper changes for overcharge.

totony-misc:
I thought the consensus was removing stuns from stones (which is done) and giving overcharge back

theQuazz:
I'm a fan

PR #735, opened 2021-01-26
Replaced Reincarnation with Anchor Soul.

Jaccouille:
$changelog: Replaced Reincarnation with Anchor Soul for both Master Healer and Sage.
$changelog: Added the Anchor Soul ability, which prevents a unit from dying momentarily.

Upon taking fatal HP damage, incoming damage will be denied and hp won't be reduced to 1, reason is, I had issue during testing with the unit affected having 1 HP and dying from the mage masher manaburn proc, didn't seem related to stat loss when I debugged.

Open to discussion.


totony-misc:
> didn't seem related to stat loss when I debugged.

might be because of normal stats degen

jonathanmclaus:
There is no normal stat degen. It's set to 0 for trolls and degen is handled by `StatLoss` package.

jonathanmclaus:
Did you go through the various instances of `subHP` (and maybe `setHP`)? For example, Sleep Outside does `subHP` without using the damage system, so I imagine this wouldn't catch it. We should switch these to damages. I'm working on that now, but just want to check if you tested it.

jonathanmclaus:
If we change Sleep Outside to use real damage, then it will interrupt its own status effect, waking the unit up immediately after invulnerability ends (under a second).

Issue #736, opened 2021-01-29
Transactions among shared gold can be incorrectly interweaved.


jonathanmclaus:
Shared gold is based around the concept of rebalancing the gold based on the tribe player after each transaction. Because gold is not immediately gained upon `EVENT_PLAYER_UNIT_PAWN_ITEM`, the rebalancing is delayed using `nullTimer`. This can lead to the following situation:

1. Player X buys an item, scheduling callback A to happen "instantaneously".
2. Player Y sells an item, scheduling callback B to happen "instantaneously".
3. Callback A is executed, which tries to syncs gold and fails because both X and Y differ from the tribe gold.
4. Callback B is executed with the same failure.

If it were possible to look up the gold for the pawned item, the gold per player could be handled via variables and not rely on the game state. These variables could be updated instantly and then use a timer to simply synce the game state to the variables for display purposes. However, this cannot be done without more scaffolding to support such a lookup.

Instead, the callback will be changed so that instead of syncing each player's gold to the one that differs in the pool, the differential between each player and the tribe gold will be applied. For example, if the tribe has 20 gold, one player has 17, and another player has 25, the net of -3 + 5 will be applied to the pool and then gold will be synced using this value.

There will also be a lock used to prevent a player from entering and leaving the pool while a callback is scheduled, in case such a trigger could be executed between steps 1 and 3 above. It may be true that only non-triggered game events like step 2 above can be interweaved, but it is simple enough to bundle together this new style of syncing and then applying any updates to the settings whenever either action is needed, meaning the callback will look like:

1. Sync gold using the new differential system.
2. Handle any pending setting changes.

And this callback will be executed both after transactions and after a setting change is desired.

jonathanmclaus:
This was implemented by separating those two steps into separate callbacks, but using a semaphore to ensure that the second step does not happen while the first step is pending.

jonathanmclaus:
Closing this ticket. I think the work is good enough that using variables to hold the gold per player is unnecessary.

PR #737, opened 2021-01-29
Fixed desynchronization of shared gold pools.

jonathanmclaus:
$changelog: Fixed bug where simultaneous transactions could desynchronize shared gold.


jonathanmclaus:
Addresses #736 partially.

PR #738, opened 2021-01-30
Added tooltip commands for testing unittypes.

jonathanmclaus:


jonathanmclaus:
Squash merge commit was deleted because it has to be reworded.

PR #750, opened 2021-02-08
Change Mana Burn on Mage Masher to be proportional to current mana.

Jaccouille:
$changelog: Changed Mage Masher to burn 5% of current mana, minimum 10.
$changelog: Fixed bug where Mage Masher could activate during some spells.

I recoded mage masher mana burn effect so we can move on this point rather than dragging the issue on discord for X month, I am starting to forget the point we had reached so I would pushed this.

I scrolled up the discord channel a bit and followed this suggestion from @theQuazz


jonathanmclaus:
The changelog says 2.5% but the code and Quazz say 5%. It's just a typo in the changelog, right? Using 2.5% means that a unit would need 400 mana before this matters.

theQuazz:
🥇

PR #751, opened 2021-02-08
Replaced raw buff id by explicit reference from stdlib

Jaccouille:
Recently updated the wurst stdlib to include a buffIds class, replaced the raw buff id, haven't tested if every buff works as intended yet.


jonathanmclaus:
Can you change this to just use `Assets`? I know your PR on stdlib is pending, but I'm fine merging this change referencing `Assets` before that.

PR #757, opened 2021-03-07
[WIP] Implemented pop up custom UI

Jaccouille:
This is a custom UI pop up frame that display 3 icons upon purchase or artifact item (horn of mammoth, pinion...) pickup to allied players.

I believe it can increase clarity/coordination and make the game a bit smoother for certain player, I am a lazy person, I don't like to have to write "got forge, bought mud hut....".

I don't think custom UI are really necessary for this but I had the idea and I think it looks neat.

### DIFF
--- a/imports/MyTableOfContent.toc
+++ b/imports/MyTableOfContent.toc
@@ -1,0 +1,1 @@
+PopUp.fdf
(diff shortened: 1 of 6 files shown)


Jaccouille:
Although this would not be a bad idea to improve clarity and information sharing in a team game, considering the current unstable Warcraft 3 environment, I believe this change would require too much time to implement.

A simple text displayed to allied players can replace this feature while being more reliable.

PR #759, opened 2021-09-13
Replaced the -as command with -dps.

Jaccouille:
$changelog: Replaced the -as command with -dps.

The command was unclear to players, they thought it gave the current attack speed while it can only give the basic attack cooldown of the unit.

It is possible to compute the attack speed, but you'd have to check for buff and items, which seem to be a hassle.

The computing agility bonus is rather easy since you can get the unit agility.

The following formula can compute the number of attacks per second: (1 + bonus attack speed in %) / base attack cooldown
[source](https://www.hiveworkshop.com/threads/attack-speed-calcuation.199711/#post-1961841)

One alternative would be to place a dummy unit and order the unit to attack it while starting a timer, but it requires the unit to be out of combat, should only be allowed in test mode.

Not really motivated to implement either of those solutions so I just updated the printed message to be clearer.


Jaccouille:
Well could also add a command to compute attack speed based on player input so he can add the bonus attack from items & abilities himself, that sounds more reasonable.

Jaccouille:
Added a -dps command that computes attack speed & DPS based on player input.




@jonathanmclaus is there any reason this command wasn't included in the quest log & the [Commands.wurst](https://github.com/island-troll-tribes/island-troll-tribes/blob/master/wurst/systems/commands/Commands.wurst) file?

Should rename the file to "DamagePerSecond.wurst" if we don't want to include it in the file linked previously.

Jaccouille:
[Source](https://www.hiveworkshop.com/threads/max-attack-speed-myth-or.86379/post-701884) describing how animation backswings point work.

wurst/systems/modes/Quests.wurst
```diff
@@ -15,6 +15,7 @@ let coolCommands = asList(
     "-rn:|r shows the tribe per player".color(HIGHLIGHT_COLOR),
     "-sd:|r Enable/disable damage display".color(HIGHLIGHT_COLOR),
     "-ms:|r displays your troll's movement speed".color(HIGHLIGHT_COLOR),
+    "-dps #attack speed #damage:|r displays your troll's attack speed & DPS".color(HIGHLIGHT_COLOR),
```
jonathanmclaus, line 18: The description for `-as` was never displayed here, so it is not in the diff even though the command is deleted.

PR #762, opened 2021-09-13
Added spell resistance to Mammoth and Disco Duck.

Jaccouille:
$changelog: Added 30% Spell Resistance to Mammoth & Disco Duck.

This change has been requested for a while as it seems too easy for mage class to deal damage on mid-bosses.

I reused the MagicResistance.wurst code made by @HallerJohn to save time & to increase the codebase.

Should replace the "magic resistance" terminology by "spell resistance", magic implies that magic damage from bees are reduced, while it is not the case, only reduce damage from spells, it uses the runed brace item ability.

### DIFF
--- a/wurst/objects/abilities/BossAbilities.wurst
+++ b/wurst/objects/abilities/BossAbilities.wurst
@@ -111,6 +111,6 @@ public let ABILITY_SPELL_DMG_REDUCTION_DUMMY_BUFF = compiletime(
 // Used by bosses
 @compiletime function createBossSpellResistance() returns AbilityDefinitionRunedBracers
     return new AbilityDefinitionRunedBracers(ABILITY_SPELL_DMG_REDUCTION)
-        ..setDamageReduction(1, 0.3)
+        ..setDamageReduction(1, SPELL_RESISTANCE)
         ..setName("Boss Spell Resistance")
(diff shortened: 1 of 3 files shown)


jonathanmclaus:
Can you add this ability to the object definition of the unit or does that not support a specific level of hero abilities?

Jaccouille:
> 
> 
> Can you add this ability to the object definition of the unit or does that not support a specific level of hero abilities?

The object definition doesn't support specific level, you can only set the level of an ability through trigger.

jonathanmclaus:
I think this should just be its own ability. Relying on the ability used for pets doesn't really make sense. If you change the increment for magic resist on pets, you'd end up changing the spell resistance on bosses too.

Jaccouille:
> 
> 
> I think this should just be its own ability. Relying on the ability used for pets doesn't really make sense. If you change the increment for magic resist on pets, you'd end up changing the spell resistance on bosses too.

Good point, I will update the PR.

Jaccouille:
I updated the mid-boss abilities definition file a bit, added dummy buffs to display an icon for spell resistance & hardened skin abilities, I think those numbers are important for the players to know. I find the implementation a bit messy though

PR #774, opened 2021-09-22
Reduce Shape Shifter bear form damage.

HallerJohn:
$changelog: Reduced Shape Shifter bear form damage from 25 -> 17.

### DIFF
--- a/wurst/objects/units/TrollUnitInitializer.wurst
+++ b/wurst/objects/units/TrollUnitInitializer.wurst
@@ -369,7 +369,7 @@ function createWolfForm(int unitId) returns HeroDefinition
 function createBearForm(int unitId) returns HeroDefinition
     return createSubTroll(unitId)
         ..setNameEnhance("Bear Form")
-        ..setAttackDamage(25)
+        ..setAttackDamage(17)
         ..setAttackSpeed(2.50)
         ..setModelFile(Units.grizzlyBear)
         ..setIcons(Icons.bTNGrizzlyBear)


Jaccouille:
It feels complicated to adjust the unit definition attributes because the agility & items in-game have a big impact on move speed & damage per second.
I ran a few tests, initially thought this damage was a bit drastic but it seems ok, I'd like to add those changes to shapeshifter considering feedback I've been reading/watching for 2 weeks :

- Bear attack speed 2.5 -> 3.0 (Attack speed was still too high from my POV with battle glove + panic)
- Bear movement speed 300 -> 210 (Bear should be slow and not used to chase/run away)
- Bear's Bulwark armor bonus 4 -> 3 (Basically mud hut aura)
- Remove Bear's Bulwark HP regeneration (Healing allies shouldn't be part of shapeshifter identity)
- Wolf attack speed 1.40 -> 1.55 (Just to nerf this form, will still have highest DPS amongst other forms)

Each form still have to feel like they're useful in a specific situation

- Wolf -> DPS
- Bear -> Tank
- Panther -> Sneak
- Tiger -> Chase

Jaccouille:
Precise the value changed in the changelog, numbers are important.

HallerJohn:
Some of the nerfs have been removed with the intent to introduce the nerfs slower over time so we can see if they are still needed or not.

jonathanmclaus:
Is the idea to just playtest this change? It's sloppy to leave the variable in otherwise.

HallerJohn:
Are you referring to `LIFE_REGENERATION_RATE`? If so I thought it was more clear to leave that at the top so that if it was changed in the future it would be easily visible.

PR #775, opened 2021-09-23
Removed mana burn visual effect from drain mana items

Jaccouille:
$changelog: Removed mana burn visual effect from dark spear & thistles

Dark spear & thistles apply a mana burn FX, this is confusing because mana burn FX implies spell damage from mana burn, those items only drain mana.


jonathanmclaus:
Doesn't this also remove the FX from items that do actually burn? Maybe you should check `burn` there?

jonathanmclaus (approved):

PR #778, opened 2022-01-04
Fixed error when building map

Jaccouille:
We added a color for the map name in #773, I didn't notice that the release script used the map name defined in the wurst build file to create a string for the filename, color code on filename crashes the wurst build script.


jonathanmclaus:
Ideally this would handle anything being a color code, not just at the beginning of the map name. Can you add a TODO comment there mentioning that to make it clear it won't guarantee it.

PR #782, opened 2022-05-16
Added healing reduction buff to Cloak of healing

Jaccouille:
$changelog: Healing Cloak last healing tick now applies Heal Reduction, subsequent healing from Cloak of healing and troll abilities are reduced by 50%

Problem: Healing Cloak is a cheap item and can be easily stacked, feedback from players mentions that stacking Healing Cloak feels kinda lame/make the game less interesting, added a diminished return on healing from Healing Cloak to reduce the efficiency of stacking.

### DIFF
--- a/wurst/objects/abilities/HealingDebuff.wurst
+++ b/wurst/objects/abilities/HealingDebuff.wurst
@@ -1,0 +1,48 @@
+package HealingDebuff
+
+// Standard library imports:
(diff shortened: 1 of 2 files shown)


jonathanmclaus:
Addressed in #788. This implementation would reduce healing from spells in addition to the cloak.

PR #786, opened 2022-09-16
Reworked spear recovery.

totony-misc:
$changelog: Changed spears to recover if and only if the target immediately dies.


totony-misc:
Poisons spears don't work when thrown via a dummy, any idea why? I'll just remove the dummies if you don't know

PR #787, opened 2022-09-16
prevent healing cloak stacking + higher cd

totony-misc:
$changelog: Healing cloak cooldown went from 35 to 50 seconds. Using two healing cloaks concurrently causes the second one to heal for 50% less.


jonathanmclaus:
This doesn't just nerf concurrent healing - it nerfs all healing for the cloak. If you use a cloak with only one tick left on the previous cast, you're getting screwed.

jonathanmclaus:
Addressed in #788

PR #789, opened 2022-09-22
Reworked ramps between NW and SW islands.

totony-misc:
the ramp was kind of in an awkward spot, maybe a tree is better?

$changelog: Removed secondary ramp between NW and SW islands.
$changelog: Changed primary ramp between NW and SW islands to be blocked with a rotten tree.


totony-misc:
image

